Issue #4, opened 2019-04-24
Is there anyway to search for specific object fields


alaviss:
I'm having an object like this:

```nim
type
  User* = object
    name {.unique.}: string
    password: string
```

Later on I'd want to get the object with `name = "username"` and there doesn't seem to be any integrated method for this. I'm I stuck with executing SQL manually via `dbConn`?

moigagoo:
Right now you can do `User.getMany(1, where="name=username")[0]`.

alaviss:
Well, but there's no way to safely compose that with user-inputted data. I guess I'd just use the `dbConn` and `to` to get the needed data for now.

I've found `nim-orm` which seems to feature a rather [nice syntax][0] for finding data, and it'd be nice if `norm` could have something like that as well.

[0]: https://github.com/nanoant/nim-orm#usage

alaviss:
Ormin even went so far as to have a sql-like dsl for this: https://github.com/Araq/blog/blob/master/ormin.rst

alaviss:
For `getMany` and friends where you can generate a `where` argument, I think a mechanism to bind the `?` to a value would be needed, as more often than not, we search within our database with input from the outside, and blindly inserting them would only results in SQL injection.

moigagoo:
I've been experimenting with different ways to increase safety in queries generated by Norm. Looking for a solution that would not change the existing interface much or be expensive to implement. I hope to share something soon.

moigagoo:
Maybe we can just allow `where` to be a tuple of a string with `?` placeholders and an array of values for them?

alaviss:
Sure. A `tuple[condition: string, params: seq[string]]` sounds easy enough to use. An alternative is to have a ``params: varargs[string, `$`]`` at the end of argument list, like those in `db_*` modules.

moigagoo:
@alaviss I've added `params` arg to `getMany` and `getOne` similar to db_* modules.

However, non-string params are not converted to string as I'd expect them:
```
# in sqlite.nim

      template getOne(obj: var object, where = "id = ?", params: varargs[string, `$`]) {.used.} =
        ##[ Read a record from DB by condition and store it into an existing object instance.

        If multiple records are found, return the first one.
        ]##

        let getOneQuery = genGetOneQuery(obj, where)

        debug getOneQuery, " <- ", params.join(", ")

        let row = dbConn.getRow(getOneQuery, params)

        if row.isEmpty():
[...]

PR #10, opened 2019-05-05
Feature/safer queries

moigagoo:
Attempt to offer a safer option to construct queries. As of Norm 1.0.8, `getMany` queries are plain strings. Any user-provided data is inserted as is or has to be prepared by your program.

With the changes in this PR, conditions for `getMany` and `getOne` procs can contain placeholders. The values for them are provided in a separate `params` argument.

The PR introduces backward incompatible changes. Instead of two args `where` and `orderBy` there's only one `cond`. If you need ordering, just put `ORDER BY` in your `cond`.

Fix https://github.com/moigagoo/norm/issues/4

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,10 +94,11 @@ Any contributions are welcome, be it pull requests, code reviews, documentation 
 If you decide to contribute through code, please run the tests after you change the code:
 
 ```shell
-$ docker-compose run tests        # run all tests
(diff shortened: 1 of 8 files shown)


moigagoo:
@alaviss I'd be grateful if you could find the time to review this PR.

Issue #3, opened 2019-03-05
Allow inserting and reading NULL values


moigagoo:
Currently, Nim's stdlib db_* modules don't support `NULL` values. There're a few working solutions, we should investigate further: https://forum.nim-lang.org/t/4693#29301

- [x] SQLite
- [ ] Postgres

JohnAD:
@moigagoo,

I've struggled with this issue in other contexts and I've come up with another awkward solution: I wrote a universal macro wrapper library called UR. But, just like using Options[T], it is kind of overkill in a lot of cases. But if curious: https://github.com/JohnAD/ur

But, I'd like to suggest a simpler answer: creating a "dynamic"-ish data type to compliment the basic datatypes. And, do this for people programming Nim in general: not just for this library. I spent a few hours creating a quick proof-of-concept at https://github.com/JohnAD/nullable   (Despite appearance, it's not a library yet.) Just look at the `nullable.nim` file to get an idea of what I'm talking about.

So, instead of this:

``` nim
db("petshop.db", "", "", ""):     # Set DB connection credentials.
  type                            # Describe object model in an ordinary type section.
    User = object
      age: int 
      name {.
        formatIt: capitalize(it)  # Guarantee that ``name`` is stored in DB capitalized.
      .}: string
```

The programmer could do this:

``` nim
db("petshop.db", "", "", ""):     # Set DB connection credentials.
  type                            # Describe object model in an ordinary type section.
    User = object
      age: Nint 
      name {.
        formatIt: capitalize(it)  # Guarantee that ``name`` is stored in DB capitalized.
      .}: Nstring
```

Now, the programmer can still do non-null, albeit in a awkward way in nim 0.19.4 **:

``` nim
withDb:                           # Start a DB session.
  createTables(force=true)        # Create tables for all objects. Drop tables if they exist.

  var bob = User(                 # Create a ``User`` instance as you normally would.
    name: parse("bob"),                  # Note that the instance is mutable. This is mandatory!
    age: parse(23)
  )
  bob.insert()  
```

The programmer can still do math, etc using the libraries overloading operators.

And now nulls could be made to work:

``` nim
withDb:                           # Start a DB session.
  createTables(force=true)        # Create tables for all objects. Drop tables if they exist.

  var bob = User(                 # Create a ``User`` instance as you normally would.
    name: Null(),                  # We are inserting a null!
    age: parse(23)
  )
  bob.insert()  
```

Thoughts?

[...]

PR #15, opened 2019-05-25
[WIP] Add support for MongoDB

JohnAD:
This is work-in-progress PR; so definitely don't accept it! It is simply so that you can watch progress as I work on adding support for MongoDB.

The first change is a tiny one: I've added object support for `Oid` from the nim standard library. MongoDB uses oids for standard identification, so is critical for basic use.

With conditional compilation on production releases, an app does does not use an oid should not see an increase in code size. I want to verify that later.

### DIFF
--- a/src/norm/rowutils.nim
+++ b/src/norm/rowutils.nim
@@ -2,10 +2,9 @@ import strutils, sequtils
 import sugar
 import macros; export macros
 import typetraits
+import oids, times
 
 import objutils, pragmas, universal
-
-import oids
 
 type Row = seq[string]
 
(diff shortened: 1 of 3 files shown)


src/norm/rowutils.nim
```diff
@@ -2,11 +2,11 @@ import strutils, sequtils
 import sugar
 import macros; export macros
 import typetraits
-import oids, times
+import oids, times, options
 
 import objutils, pragmas, universal
 
-type Row = seq[string]
+type Row = seq[Option[string]]
```
moigagoo: @JohnAD not really sure this is a good idea... We're fundamentally changing the way Rows are treated, breaking compat with db_* modules, rewriting all code that handles Rows, all for the _optional_ feature of null checks...

However, as a temporary hack it's OK as far as it helps you move forward. I'd love to see the complete PR, maybe it'll all make sense in the right context.

JohnAD: I will be curious as to how they fix the db_* libraries. Since `nil` is being deprecated they need to find another way to handle nulls. For now, this will work, sort of, for MongoDB dev.

Whatever they come up with, I agree it's best to match their new null-handling technique.

JohnAD:
As mentioned in #3 I'm closing this PR. I'll starting a new PR based on `develop` branch later.

PR #17, opened 2019-06-13
Add `withCustomDb` to allow running commands under custom DB connections; add NOT NULL constraint by default for non-Option non-custom types

moigagoo:


JohnAD:
I like it. Here is an idea to make it look cleaner. Rather than have the four parameters, you could have it passed in as an object.

so,

```nim
other_db = dbConnect("postgres", "postgres", "", "postgress")  # returns an object

withCustomDb(other_db):
  createTables(force=true)
```

it could also be invoked directly:

```nim
withCustomDb(dbConnect("postgres", "postgres", "", "postgress")):
  createTables(force=true)
```

JohnAD:
Or it could be made universal:

```nim
main_db = dbParameters("postgres", "postgres", "", "postgress")
other_db = dbParameters("postgres2", "postgres", "", "postgress")

db(main_db):
  type
    User = object
      age: int
      name {.
        formatIt: capitalize(it)
      .}: string

withDb:
  createTables(force=true)
withDb(other_db):  # two templates with same name for compile-time detection 
  createTables(force=true)
```

moigagoo:
I tried that first. There's a problem with this solution: you can't reopen closed connections. In fact, I really wanted to pass DbConn to db when I first wrote it. I also wanted there to be a way to reuse persistent connections. But that turned out to be unsafe and inconvenient.

Установить Outlook для iOS<https://aka.ms/o0ukef>
________________________________
От: John Dupuy <notifications@github.com>
Отправлено: Thursday, June 13, 2019 7:07:43 PM
Кому: moigagoo/norm
Копия: Constantine Molchanov; Author
Тема: Re: [moigagoo/norm] Add `withCustomDb` to allow running commands under custom DB connections. (#17)


I like it. Here is an idea to make it look cleaner. Rather than have the four parameters, you could have it passed in as an object.

so,

other_db = dbConnect("postgres", "postgres", "", "postgress")  # returns an object

withCustomDb(other_db):
  createTables(force=true)

it could also be invoked directly:

withCustomDb(dbConnect("postgres", "postgres", "", "postgress")):
  createTables(force=true)

JohnAD:
Ah, I didn't know that. I'm might play around with it some more then.

moigagoo:
I'm going to merge this one if you approve of the code despite the design decision (I promise, it's a necessary evil 😊).

Установить Outlook для iOS<https://aka.ms/o0ukef>
________________________________
От: John Dupuy <notifications@github.com>
Отправлено: Thursday, June 13, 2019 7:24:53 PM
Кому: moigagoo/norm
Копия: Constantine Molchanov; Author
Тема: Re: [moigagoo/norm] Add `withCustomDb` to allow running commands under custom DB connections. (#17)


Ah, I didn't know that. I'm might play around with it some more then.

PR #18, opened 2019-06-16
Added support for databaseParameters proc.

JohnAD:
I have learned **much** more about template parameters in the last four hours :).

Anyway, I have parameter objects working if you are interested. This is a non-breaking change, it is simply an expansion of how the `db` macro (and subtending `withCustomDb` templates) can behave.

Example of use:

```nim
# first, verifying that the previous methods still work:

db("testA.db", "", "", ""):
  type
    User* = object
      display_name*: string
      zip_code*: Option[string]

var u = User(
  display_name: "joe3",
  zip_code: none string
)

withDb:
  createTables(force=true)
  u.insert()

withCustomDb("testB.db", "", "", ""):
  createTables(force=true)
  u.insert()

# now the new stuff:

let dbC = databaseParameters("testC.db", "", "", "")

withCustomDb(dbC):
  createTables(force=true)
  u.insert()

let dbD = databaseParameters("testD.db", "", "", "")

db(dbD):
  type
    Animal* = object
      species: string
      weight: int

var a = Animal(species: "cat")

withDb:  # this writes to "testD.db" as seen in dbD
  createTables(force=true)
  a.insert()

withCustomDb(dbC):
  createTables(force=true) # this will not touch the existing 'user' table
  a.insert()

withCustomDb(databaseParameters("testE.db", "", "", "")):
  createTables(force=true)
  a.insert()
```

### DIFF
--- a/src/norm/sqlite.nim
+++ b/src/norm/sqlite.nim
@@ -18,6 +18,19 @@ export sqlite
 export rowutils, objutils, pragmas
 
 
+type
+  DbParamFields* = object
+    stored_connection*: string 
+    stored_user*: string
+    stored_password*: string
+    stored_database*: string
+
+proc databaseParameters*(connection, user, password, database: string): DbParamFields =
+  result.stored_connection = connection
+  result.stored_user = user
+  result.stored_password = password
+  result.stored_database = database
+
 proc `$`*(query: SqlQuery): string = $ string query
 
 proc getTable*(objRepr: ObjRepr): string =
@@ -181,7 +194,7 @@ template genWithDb(connection, user, password, database: string,
                    tableSchemas, dropTableQueries: openArray[SqlQuery]): untyped {.dirty.} =
   ## Generate ``withDb`` templates.
 
-  template withCustomDb*(customConnection, customUser, customPassword, customDatabase: string,
+  template withSpecificDb(customConnection, customUser, customPassword, customDatabase: string,
                          body: untyped): untyped {.dirty.} =
     ##[ A wrapper for actions that require custom DB connection, i.e. not the one defined in ``db``.
     Defines CRUD procs to work with the DB, as well as ``createTables`` and ``dropTables`` procs.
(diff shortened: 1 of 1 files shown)


moigagoo:
What's the goal though? If the goal is to allow grouping db connection params under aliases, then there's a simpler and shorter solution: 

```diff
diff --git a/src/norm/sqlite.nim b/src/norm/sqlite.nim
index 35a1070..b2aae15 100644
--- a/src/norm/sqlite.nim
+++ b/src/norm/sqlite.nim
@@ -18,6 +18,9 @@ export sqlite
 export rowutils, objutils, pragmas
 
 
+type DbParams* = tuple[connection, user, password, database: string]
+
+
 proc `$`*(query: SqlQuery): string = $ string query
 
 proc getTable*(objRepr: ObjRepr): string =
@@ -337,6 +340,10 @@ template genWithDb(connection, user, password, database: string,
         body
       finally: dbConn.close()
 
+  template withCustomDb*(dbParams: DbParams, body: untyped): untyped {.dirty.} =
+    withCustomDb(dbParams.connection, dbParams.user, dbParams.password, dbParams.database):
+      body
+
   template withDb*(body: untyped): untyped {.dirty.} =
     ##[ A wrapper for actions that require DB connection. Defines CRUD procs to work with the DB,
     as well as ``createTables`` and ``dropTables`` procs.
@@ -402,3 +409,7 @@ macro db*(connection, user, password, database: string, body: untyped): untyped
                                     genTableSchemas(dbObjReprs), genDropTableQueries(dbObjReprs))
 
   result.insert(0, withDbNode)
+
+template db*(dbParams: DbParams, body: untyped): untyped =
+  db(dbParams.connection, dbParams.user, dbParams.password, dbParams.database):
+    body
```

I think such aliases could be useful if we could pass several of them at the same time, e.g. `db(master, replica)`. We could use threads to run the commands on multiple DBs in parallel.

But whether we need to implement such a feature, is an open question.

JohnAD:
Using a template for the alternate `db` call makes a lot of a sense. Good catch.

A tuple would work.

In theory, one could avoid adding a new element at all by using an array:

```nim
template db*(p: array[0..3, string], body: untyped): untyped =
  db(p[0], p[1], p[2], p[3]):
    body
```

Though in practice, I've not debugged that. For some reason the `body` parameter is being eaten.

Anyway, the goal:

I was thinking long-term, as `norm` moves into production, an object would be more flexible. Allowing optional parameters, non-string parameters, meta information, and so on.

moigagoo:
@JohnAD I always forget to click "Submit Review" after reviewing PRs :-( I've left some comments on the code.

However, my main concern is not the quality of the submitted code, but the necessity of this feature at this moment in Norm's development. The only benefit I see in DB param aliases so far is that it allows to avoid repeating the same params in `withCustomDb` calls. However, this doesn't feel like a common case. So I'm feeling this is better implemented after we:
- move postgres to ndb
- factor repeating functionality in sqlite and postgres to a shared module
- possibly refactor rowutils because it currently contains logic that is maybe better suited for specific backends
- find a better way to determine field type that `$fieldRepr.type == "int"` (man, I hate this code so much!)
- evaluate the importance of supporting multidb setups (like the master-replica example).

All in all, there's so much fundamental issues to resolve that I feel that this is a premature optimization. But this is just my stand on the thing, I'm open to discussion.

moigagoo:
> In theory, one could avoid adding a new element at all by using an array:

But the calls would be really ugly: `db(["foo.db", "", '", ""]):`. So many brackets!

JohnAD:
> All in all, there's so much fundamental issues to resolve that I feel that this is a premature optimization. But this is just my stand on the thing, I'm open to discussion.

I can see that. My head is in the MongoDB branch anyway right now.

PR #19, opened 2019-06-19
[WIP] Adding MongoDB support

JohnAD:
PR just for tracking purposes; please don't merge.

Start of the process. Right now, it just barely works and is filled with bugs. You can connect to a MongoDB server and insert a single document; as long as that document/object only has strings and OIds in it. :)

Much more to do.

### DIFF
--- a/src/norm/mongodb.nim
+++ b/src/norm/mongodb.nim
@@ -139,6 +139,8 @@ proc getCollectionName*(objRepr: ObjRepr): string =
     if prag.name == "table" and prag.kind == pkKval:
       return $prag.value
 
+proc getTableName*(objRepr: ObjRepr): string =
+  getCollectionName(objRepr)
 
 proc getCollectionName*(T: typedesc): string =
   ##[ Get the name of the DB table for the given type: ``table`` pragma value if it exists
@@ -148,6 +150,8 @@ proc getCollectionName*(T: typedesc): string =
   when T.hasCustomPragma(table): T.getCustomPragmaVal(table)
   else: ($T).toLowerAscii()  
 
+proc getTableName*(T: typedesc): string =
+  getCollectionName(T)
 
 var
   allCollections*: seq[string] = @[]
@@ -156,7 +160,7 @@ var
   normPassword*: string
   normDatabase*: string
 
-template genWithoutDb(newCollections: seq[string]): untyped {.dirty.} =
+template addObjectsToCollection(newCollections: seq[string]): untyped {.dirty.} =
   ## simply add to global allCollections
 
   allCollections.insert(newCollections)
@@ -220,29 +224,28 @@ template genWithDb(connection, user, password, database: string, useParams: int,
         let
           doc = toBson(obj, force)
           dbCollection = dbConn[customDatabase][getCollectionName(type(obj))]
+          response = dbCollection.insert(doc)
 
-        var response = dbCollection.insert(doc)
-
-        if len(response.inserted_ids) > 0:
+        if len(response.inserted_ids) == 1:  # TODO: handle or document not-1 failure scenarios
           obj.id = response.inserted_ids[0].toOid
 
-      proc pullOne(obj: var object, id: Oid) {.used.} =
+      proc getOne(obj: var object, id: Oid) {.used.} =
         ## Read a record from DB by id and apply it to an existing object instance.
 
         let
           dbCollection = dbConn[customDatabase][getCollectionName(type(obj))]
 
         let fetched = dbCollection.find(%*{"_id": id}).one()
-        applyBSON(obj, fetched)
+        applyBson(obj, fetched)
 
-      proc pullOne(obj: var object, cond: Bson) {.used.} =
+      proc getOne(obj: var object, cond: Bson) {.used.} =
         ## Read a record from DB by search condition and apply it into an existing object instance.
 
         let
           dbCollection = dbConn[customDatabase][getCollectionName(type(obj))]
 
         let fetched = dbCollection.find(cond).one()
-        applyBSON(obj, fetched)
+        applyBson(obj, fetched)
 
       proc getOne(T: typedesc, id: Oid): T {.used.} =
         ## Read a record from DB by id and return a new object
@@ -251,7 +254,7 @@ template genWithDb(connection, user, password, database: string, useParams: int,
           dbCollection = dbConn[customDatabase][getCollectionName(T)]
 
         let fetched = dbCollection.find(%*{"_id": id}).one()
-        applyBSON(result, fetched)
+        applyBson(result, fetched)
 
       proc getOne(T: typedesc, cond: Bson): T {.used.} =
         ## Read a record from DB by search condition and return a new object
@@ -260,7 +263,7 @@ template genWithDb(connection, user, password, database: string, useParams: int,
           dbCollection = dbConn[customDatabase][getCollectionName(T)]
 
         let fetched = dbCollection.find(cond).one()
-        applyBSON(result, fetched)
+        applyBson(result, fetched)
 
       proc getMany(
         T: typedesc,
@@ -277,10 +280,10 @@ template genWithDb(connection, user, password, database: string, useParams: int,
         let fetched = cursor.all()
         for doc in fetched:
           var temp = T()
-          applyBSON(temp, doc)
+          applyBson(temp, doc)
           result.add temp
 
-      proc pullMany(
+      proc getMany(
         objs: var seq[object],
         nLimit: int,
         offset = 0,
@@ -359,6 +362,11 @@ proc ensureIdFields(typeSection: NimNode): NimNode =
     result.add objRepr.toTypeDef()
 
 proc reconstructType(n: NimNode): string =
+  ##[
+    For the node passed in, generate a normalized string representation of the type.
+
+    This function handles both plain types as well as compound types such as seq[T] and Option[T].
+  ]##
   if n.kind == nnkIdent:
     return $n
   if n.kind == nnkBracketExpr:
@@ -368,13 +376,19 @@ proc reconstructType(n: NimNode): string =
   return "unknown"
 
 proc updateObjectRegistry(dbObjReprs: seq[ObjRepr]) {.compileTime.} =
+  ##[
+    Add a new entry in into the global variable `normObjectNamesRegistry`.
+
+    This registry only exists at compile-time and is used by various
+    macros.
+  ]##
   var objName: string
   for obj in dbObjReprs:
     objName = obj.signature.name
     if not normObjectNamesRegistry.contains(objName):
       normObjectNamesRegistry.add objName
 
-proc genObjectToBSON(dbObjReprs: seq[ObjRepr]): string =
+proc genObjectToBson(dbObjReprs: seq[ObjRepr]): string =
   ##[
   this procedure generates new procedures the convert the values in an
   existing "type" object to a BSON object.
@@ -478,28 +492,28 @@ proc genObjectToBSON(dbObjReprs: seq[ObjRepr]): string =
     result &= s
     result &= "\n" # add a blank line between each proc
 
-const TYPE_TO_BSON_KIND = {
-  "float": "BsonKindDouble",
-  "string": "BsonKindStringUTF8",
-  "Oid": "BsonKindOid",
-  "bool": "BsonKindBool",
-  "Time": "BsonKindTimeUTC",
-  "int": "BsonKindInt64, BsonKindInt32"
-}.toTable
+const
+  TYPE_TO_BSON_KIND = {
+    "float": "BsonKindDouble",
+    "string": "BsonKindStringUTF8",
+    "Oid": "BsonKindOid",
+    "bool": "BsonKindBool",
+    "Time": "BsonKindTimeUTC",
+    "int": "BsonKindInt64, BsonKindInt32"
+  }.toTable
+  TYPE_TO_BSON_PROC = {
+    "float": "toFloat64",
+    "string": "toString",
+    "Oid": "toOid",
+    "bool": "toBool",
+    "Time": "toTime",
+    "int": "toInt"
+  }.toTable
 
-const TYPE_TO_BSON_PROC = {
-  "float": "toFloat64",
-  "string": "toString",
-  "Oid": "toOid",
-  "bool": "toBool",
-  "Time": "toTime",
-  "int": "toInt"
-}.toTable
-
-proc genBSONToObject(dbObjReprs: seq[ObjRepr]): string =
+proc genBsonToObject(dbObjReprs: seq[ObjRepr]): string =
   ##[
   this procedure generates new procedures that map values found in an
-  existing "type" object to a BSON object.
+  existing "type" object to a Bson object.
 
   So, for example, with object defined as:
 
@@ -516,12 +530,12 @@ proc genBSONToObject(dbObjReprs: seq[ObjRepr]): string =
   you will get a string containing procedures similar to:
 
   ```
-  proc applyBSON(obj: var Pet, doc: Bson) =
+  proc applyBson(obj: var Pet, doc: Bson) =
     if doc.contains("shortName"):
       if doc["shortName"].kind in @[BsonKindStringUTF8]:
         obj.shortName = doc["shortName"].toString
 
-  proc applyBSON(obj: var User, doc: Bson) =
+  proc applyBson(obj: var User, doc: Bson) =
     if doc.contains("_id"):
       if doc["_id"].kind in @[BsonKindOid]:
         obj.id = doc["_id"].toOid
@@ -533,7 +547,7 @@ proc genBSONToObject(dbObjReprs: seq[ObjRepr]): string =
         obj.display_name = doc["displayName"].toString
     if doc.contains("thePet"):
       obj.thePet = Pet()
-      applyBSON(obj.my_pet, doc["thePet"])
+      applyBson(obj.my_pet, doc["thePet"])
 
   ```
   ]##
@@ -547,12 +561,12 @@ proc genBSONToObject(dbObjReprs: seq[ObjRepr]): string =
     center = ""
 
   #
-  # now generate one applyBSON per object
+  # now generate one applyBson per object
   #
   for obj in dbObjReprs:
     objectName = obj.signature.name
     key = objectName
-    proc_map[key] =  "proc applyBSON(obj: var $1, doc: Bson) {.used.} =\n".format(objectName)
+    proc_map[key] =  "proc applyBson(obj: var $1, doc: Bson) {.used.} =\n".format(objectName)
     proc_map[key] &=  "  discard\n" # just in case there are no valid fields
     #
     # handle universal types first
@@ -597,7 +611,7 @@ proc genBSONToObject(dbObjReprs: seq[ObjRepr]): string =
       if normObjectNamesRegistry.contains(typeName):
         proc_map[key] &= "  if doc.contains(\"$1\"):\n".format(fieldName)
         proc_map[key] &= "    obj.$1 = $2()\n".format(fieldName, typeName)
-        proc_map[key] &= "    applyBSON(obj.$1, doc[\"$1\"])\n".format(fieldName)
+        proc_map[key] &= "    applyBson(obj.$1, doc[\"$1\"])\n".format(fieldName)
   #
   # finish up all procedure strings
   #
@@ -639,8 +653,8 @@ macro db*(connection, user, password, database: string, body: untyped): untyped 
 
   # echo $dbObjReprs
   # echo $genObjectAccess(dbObjReprs)
-  # echo $genObjectToBSON(dbObjReprs)
-  # echo $genBSONToObject(dbObjReprs)
+  # echo $genObjectToBson(dbObjReprs)
+  # echo $genBsonToObject(dbObjReprs)
 
 
   # let objectAccess =  parseStmt(genObjectAccess(dbObjReprs))
@@ -649,13 +663,13 @@ macro db*(connection, user, password, database: string, body: untyped): untyped 
   let withDbNode = getAst genWithDb(connection, user, password, database, -1, newCollections)  # -1 = true
   result.insert(0, withDbNode)
 
-  let bsonToObject = parseStmt(genBSONToObject(dbObjReprs))
+  let bsonToObject = parseStmt(genBsonToObject(dbObjReprs))
   result.add(bsonToObject)
 
-  let objectToBSON = parseStmt(genObjectToBSON(dbObjReprs))
-  result.add(objectToBSON)
+  let objectToBson = parseStmt(genObjectToBson(dbObjReprs))
+  result.add(objectToBson)
 
-macro dbAddCollection*(obj: typed): untyped =
+macro dbAddTable*(obj: typed): untyped =
   ##[
     Add a DB models for an object that has already been defined.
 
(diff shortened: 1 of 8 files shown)


moigagoo:
When you feel confident to have the PR reviewed, please request a review.

JohnAD:
Very close to finished. Next, I'll be writing documentation and expanding the docker file for mongodb testing.

src/norm/objutils.nim
```diff
@@ -77,15 +77,19 @@ proc toSignatureRepr(def: NimNode): SignatureRepr =
       result.name = $def[0][1]
       result.exported = true
     of nnkPragmaExpr:
-      expectKind(def[0][0], {nnkIdent, nnkPostfix})
+      expectKind(def[0][0], {nnkIdent, nnkSym, nnkPostfix})
       case def[0][0].kind
       of nnkIdent:
         result.name = $def[0][0]
+      of nnkSym:
```
moigagoo: Why not just modify the condition above to `of nnkIdent, nnkSym`?

Also, could you please explain what the difference between symbol and identifier is?

tests/testmongodb.nim
```diff
@@ -0,0 +1,219 @@
...
+const
+  # for local testing, modify your /etc/hosts file to contain "mongodb_1"
+  # pointing to your local mongodb server
+  dbConnection = "mongodb://mongodb_1:27017"
+  dbName = "TestDb"
+  customDbName = "TestCustomDb"
+
+type
+  Copyright = object
+    legal_name: string
```
moigagoo: Pleas stick to camelCase for codebase consistency.

tests/testmongodb.nim
```diff
@@ -0,0 +1,219 @@
...
+      ssn: Option[int]
+      birthDate: Time
+    Publisher {.table: "publishers".} = object
+      title {.unique.}: string
+    Book {.table: "books".} = object
+      title: string
+      authorEmail {.fk: User.email, onDelete: "CASCADE".}: string
+      publisherTitle {.fk: Publisher.title.}: string
+      ratings: seq[float]
+      copyright: Copyright
```
moigagoo: I never tested fields of custom object types defined outside `db` scope. Does it work? If so, I think the test suite for SQLite and Postgres should have this test too.

tests/testmongodb.nim
```diff
@@ -0,0 +1,219 @@
...
+
+  # TODO proc getBookById(id: Oid): Book = withDb(Book.getOne id)
+  # TODO add and test foreign key pulls in mongo
+
+type
```
moigagoo: Looks like a typo. This should be indented under `db` scope.

moigagoo: Or maybe it isn't. I see you've added `dbAddCollection` which apparently allows to add table definitions separately.

tests/testmongodb.nim
```diff
@@ -0,0 +1,219 @@
...
+suite "Creating and dropping tables, CRUD":
+  setup:
+    var 
+      user_reference_id: array[1..9, Oid]
+      publisher_reference_id: array[1..9, Oid]
+      book_reference_id: array[1..9, Oid]
+      edition_reference_id: array[1..9, Oid]
+
+    withDb:
+      createTables(force=true)
```
moigagoo: I wonder if this will cause confusion... Mongo doesn't have “tables,” it has “collections.” Maybe, introduce an alias `createCollections`?

tests/testmongodb.nim
```diff
@@ -0,0 +1,219 @@
...
+  test "Read records":
+    withDb:
+      var
+        users: seq[User] = @[]
+        # users = User().repeat 10
+        publishers = Publisher().repeat 10
+        books = Book().repeat 10
+        editions = Edition().repeat 10
+
+      users.pullMany(20, offset=5, sort = %*{"ssn": 1})
```
moigagoo: What's the deal with renaming `getMany` to `pullMany`? Is it a Mongo-specific thing? IMO it's confusing to have different terminology for essentially the same action in different backends.

tests/testmongodb.nim.cfg
```diff
@@ -0,0 +1 @@
+-d:nimOldCaseObjects
```
moigagoo: Can we switch to the new case objects and eliminate this switch?

moigagoo: Also, can we use .nims for configuration?

moigagoo:
My God, the PR is huge! I've reviewed everything but the hardest part, the 700+ line mongodb.nim monster. Got to allocate some time for that :-)

src/norm/mongodb.nim
```diff
@@ -0,0 +1,727 @@
...
+###############
+
+]##
+
+#
+# Interpreting between BSON and Nim Objects
+#
+# From BSON to Nim Types:
+#
+# | BSON          | Type      | Notes
```
moigagoo: I ❤ this table! We need that at the top of every other backend.

src/norm/mongodb.nim
```diff
@@ -0,0 +1,727 @@
...
+  "Option[int]",
+  "seq[float]",
+  "seq[string]",
+  "seq[Oid]",
+  "seq[bool]",
+  "seq[Time]",
+  "seq[int]"
+]
+
+proc getCollectionName*(objRepr: ObjRepr): string =
```
moigagoo: I can't decide what the right solution is. On one hand, there are no tables in Mongo, there are collections. Thus, it's logical to have `collection` pragma. On the other hand, since this is part of the bigger package, we want things consistent and similarly named, e.g. `table` and `getTableName` even if it's not technically a table.

src/norm/mongodb.nim
```diff
@@ -0,0 +1,727 @@
...
+  or lowercased type name otherwise.
+  ]##
+
+  when T.hasCustomPragma(table): T.getCustomPragmaVal(table)
+  else: ($T).toLowerAscii()  
+
+
+var
+  allCollections*: seq[string] = @[]
+  normConnection*: string
```
moigagoo: Why prefix with `norm`?

moigagoo: Also, why not declare all four variables on the same line?

src/norm/mongodb.nim
```diff
@@ -0,0 +1,727 @@
...
+  normUser*: string
+  normPassword*: string
+  normDatabase*: string
+
+template genWithoutDb(newCollections: seq[string]): untyped {.dirty.} =
```
moigagoo: What is the purpose of this “withoutDb” thing*

moigagoo: In the original backends, `getWithDb` stands for “generate `withDb` macro code.” So this must be generating `withoutDb` macro? But it does something completely different, I can't quite see what exactly now.

src/norm/mongodb.nim
```diff
@@ -0,0 +1,727 @@
...
+  ## simply add to global allCollections
+
+  allCollections.insert(newCollections)
+
+template genWithDb(connection, user, password, database: string, useParams: int, newCollections: seq[string]): untyped {.dirty.} =
+  ## Generate ``withDb`` templates.
+
+  allCollections.insert(newCollections)
+  if useParams == -1:
```
moigagoo: The title implies it's a bool flag (use of don't use params). But it's not, and it's not immediately clear what it means. The number of params? But why can it be -1?

moigagoo: Also, this -1 number looks like a magic number out of nowhere.

src/norm/mongodb.nim
```diff
@@ -0,0 +1,727 @@
...
+    'database' should contain the database shard.
+
+    Aforementioned procs and procs defined in a ``db`` block can be used only
+    in  a ``withDb`` block.
+    ]##
+
+    block:
+      var dbConn = newMongoWithURI(customConnection)
+      let
+        dbConnResult = dbConn.connect()  ## TODO: Add more error handling
```
moigagoo: I think, wrapping it in a try will do.

moigagoo: Maybe, it's not even necessary.

src/norm/mongodb.nim
```diff
@@ -0,0 +1,727 @@
...
+
+      template dropTables() {.used.} =
+        ## Drops the collections for ALL objects.
+
+        for c in allCollections:
+          let dbCollection = dbConn[customDatabase][c]
+          let removeResult = dbCollection.drop()
```
moigagoo: What's the purpose of `removeResult`? It's not used anywhere. A `discard` will do.

src/norm/mongodb.nim
```diff
@@ -0,0 +1,727 @@
...
+
+      template createTables(force = false) {.used.} =
+        ##[ For other database types, this function create new tables. However,
+        for MongoDB this has less meaning since collections are auto-created.
+        However, if `force` is set to true, it will dropTables first, deleting all
+        the documents.
+        ]##
+
+        if force:
```
moigagoo: Maybe put a couple of debug log statements here, similar to how SQL queries are logged in other backends? Like, “Collections are created automatically, skipping.”

src/norm/mongodb.nim
```diff
@@ -0,0 +1,727 @@
...
+        the insertion.
+
+        The ``force`` parameter is ignored by the mongodb library.
+        ]##
+
+        let
+          doc = toBson(obj, force)
+          dbCollection = dbConn[customDatabase][getCollectionName(type(obj))]
+
+        var response = dbCollection.insert(doc)
```
moigagoo: Why `var`? It's not modified downstream.

src/norm/mongodb.nim
```diff
@@ -0,0 +1,727 @@
...
+
+        if len(response.inserted_ids) > 0:
```
moigagoo: Is it possible to get several ids on one insertion? Sounds like an error. I think this condition could be made more strict, like checking if there's exactly one id (OK) or not (zero or more than one, not OK).

src/norm/mongodb.nim
```diff
@@ -0,0 +1,727 @@
...
+          obj.id = response.inserted_ids[0].toOid
+
+      proc pullOne(obj: var object, id: Oid) {.used.} =
```
moigagoo: What is the purpose o renaming getOne to pullOne? The original idea is that you have a getOne that you can use with both instances and types, which, I thought, was convenient,

moigagoo: I can see how that makes things more explicit, but I'd like to hear your reasoning.

src/norm/mongodb.nim
```diff
@@ -0,0 +1,727 @@
...
+            PragmaRepr(name: "dbCol", kind: pkKval, value: parseExpr("\"_id\""))
+          ]
+        ),
+        typ: ident "Oid"
+      )
+      objRepr.fields.insert(idField, 0)
+
+    result.add objRepr.toTypeDef()
+
+proc reconstructType(n: NimNode): string =
```
moigagoo: Please add docstrings to all procs. In Nim, procs that operate on NimNodes are extremely hard to read, so a doc is a must have.

src/norm/mongodb.nim
```diff
@@ -0,0 +1,727 @@
...
+  return "unknown"
+
+proc updateObjectRegistry(dbObjReprs: seq[ObjRepr]) {.compileTime.} =
+  var objName: string
+  for obj in dbObjReprs:
+    objName = obj.signature.name
+    if not normObjectNamesRegistry.contains(objName):
+      normObjectNamesRegistry.add objName
+
+proc genObjectToBSON(dbObjReprs: seq[ObjRepr]): string =
```
moigagoo: Nim's style guide suggests Bson instead of BSON.

src/norm/mongodb.nim
```diff
@@ -0,0 +1,727 @@
...
+          "In object $1, the field $2's type is not known to norm/mongodb. If it is a subtending object, is $3 defined by dB, dbAddCollection, or dbAddObject yet?".format(objectName, fieldName, typeName)
+        )
+  #
+  # finish up all procedure strings
+  #
+  for key, s in proc_map.pairs():
+    result &= s
+    result &= "\n" # add a blank line between each proc
+
+const TYPE_TO_BSON_KIND = {
```
moigagoo: Constants should live next to imports, at the very top.

src/norm/mongodb.nim
```diff
@@ -0,0 +1,727 @@
...
+  "float": "BsonKindDouble",
+  "string": "BsonKindStringUTF8",
+  "Oid": "BsonKindOid",
+  "bool": "BsonKindBool",
+  "Time": "BsonKindTimeUTC",
+  "int": "BsonKindInt64, BsonKindInt32"
+}.toTable
+
+const TYPE_TO_BSON_PROC = {
```
moigagoo: Two const blocks should be one.

src/norm/mongodb.nim
```diff
@@ -0,0 +1,727 @@
...
+  let withDbNode = getAst genWithDb(connection, user, password, database, -1, newCollections)  # -1 = true
+  result.insert(0, withDbNode)
+
+  let bsonToObject = parseStmt(genBSONToObject(dbObjReprs))
+  result.add(bsonToObject)
+
+  let objectToBSON = parseStmt(genObjectToBSON(dbObjReprs))
+  result.add(objectToBSON)
+
+macro dbAddCollection*(obj: typed): untyped =
```
moigagoo: I like this idea, I can see how it can be useful. Probably, should be backported to other backends.

src/norm/mongodb.nim
```diff
@@ -0,0 +1,727 @@
...
+
+  let bsonToObject = parseStmt(genBSONToObject(dbObjReprs))
+  result.add(bsonToObject)
+
+  let objectToBSON = parseStmt(genObjectToBSON(dbObjReprs))
+  result.add(objectToBSON)
+
+macro dbAddObject*(obj: typed): untyped =
+  ##[
+    Add an object for use a subtending JSON object. This macro does NOT create
```
moigagoo: Sorry, I can't quite understand what you're saying here. Maybe, add an example?

src/norm/mongodb.nim
JohnAD: For reasons I don't fully understand, Nim turns template bool parameters into ints dynamically, breaking the code. Very frustrating. I would prefer it to be a bool. Anyway, a -1 is all binary ones, so I'm using it to represent True and a zero as false. I'll add a comment to explain that. I might also submit a bug report for Nim.

src/norm/mongodb.nim
JohnAD: Ah, misunderstood the meaning. I thought it meant "with DB parameters". I will rename it. This is for objects that are not tables/collections but are part of a document (or map column if postgresql).

src/norm/mongodb.nim
JohnAD: These are essentially globals and are imported into the library users namespace. Prefixing the name with the library name makes collision much less likely IMO. I can see a programmer wanting to use a variable called "user" for example.

JohnAD: I think declaring the vars separately makes it easier to read. But I'm happy to change it.

src/norm/mongodb.nim
JohnAD: I'll make an alias. That makes sense.

src/norm/objutils.nim
JohnAD: Making the change to the case statement.

When using use the `db` macro, the name of the field is given in the source code, a `nnkIndent`. When using dbAddObject or dbAddCollection, the object already exist, so the field is a referenced symbol, a `nnkSym`. With this change, both the direct definition and the indirect reference works in `toObjRepr`.

tests/testmongodb.nim
JohnAD: making the change

tests/testmongodb.nim
JohnAD: I would not work (or should not work) for SQLite as that database cannot store map columns. But expanding Postgres to have this makes sense. I think they are called composite types in Postgres.

We would need to add the "dbAddObject" macro as well.

tests/testmongodb.nim
JohnAD: One of the benefits is that objects from other libraries can also be stored now; whereas before you could only add objects you had control over.

tests/testmongodb.nim
JohnAD: I will do that.

src/norm/mongodb.nim
JohnAD: Will change.

tests/testmongodb.nim
JohnAD: I'll change it back. Just an opinion, but that one getMany alters the first parameter and the other getMany does not is a point of confusion, but backward compatibility is important also. I will back the change.

tests/testmongodb.nim.cfg
JohnAD: I sent a message to the `nimongo` repo requesting the few simple changes needed to make `nimongo` ready for nim 0.20. Not heard back yet.

src/norm/mongodb.nim
JohnAD: Turns out to be a known bug: https://forum.nim-lang.org/t/4995

src/norm/mongodb.nim
JohnAD: I don't think the wrapping is needed but will verify. Mostly I need to the think through the failure scenarios and make sure it is documented and handled well. It won't the next commit, but it is part of my list of things to finish.

src/norm/mongodb.nim
JohnAD: That might be. I'll be documenting failure scenarios soon. If not used in the end, I will definitely discard it.

src/norm/mongodb.nim
JohnAD: Good idea. I will look into that with the failure scenario work soon.

src/norm/mongodb.nim
JohnAD: Good catch. I will make the change.

src/norm/mongodb.nim
JohnAD: It is possible MongoDB in general, but never for this function. I will add more code in when I get to handling failure scenarios step next.

src/norm/mongodb.nim
JohnAD: As mentioned in a earlier reply, I found the same function having different effects on the first parameter confusing. But I think it is better to be the same as the other databases, so I am reverting it all back to "get".

src/norm/mongodb.nim
JohnAD: Adding it now.

src/norm/mongodb.nim
JohnAD: These constants only have meaning in the context of the procedure they precede. I'll move them if you really want it though.

src/norm/mongodb.nim
JohnAD: Changing it now.

src/norm/mongodb.nim
JohnAD: I can help with that.

There is one gotcha to it that I will need both document and write a new Exception for: the Object must *already* have an 'id' field. Since the type already exists, it is not possible to add it dynamically. This needs to be caught at compile-time, so I will add a compile-time check.

src/norm/mongodb.nim
JohnAD: Adding more documentation now.

JohnAD:
I've done most of the changes.

Next I will be going over the many possible database failure scenarios and making sure they are handled gracefully.

After that, I will start reference and user documentation.

src/norm/mongodb.nim
JohnAD: They have been moved to the top and are now compile-time variables.

tests/testmongodb.nim.cfg
JohnAD: Sent another message and included another developer. Hopefully something happens soon.

Needing to use the the switch is silly given how easy it is to use the new case object methods. Plus nim 1.0 is coming soon and it would be good to have `nimongo` ready for it.

src/norm/mongodb.nim
JohnAD: Looks like there is a PR for Nim, but it is not in the release candidate. It might be a while before we can use bool template parameters safely.

src/norm/mongodb.nim
JohnAD: Error handling support has been added. Make a note to myself to remove the comment on the next commit.

JohnAD:
Latest update

Just got back from a week-long vacation, so ready to start up again. I've got a few more things to do:

* implement the remaining requested changes
* write documentation for this module

But, before I do that, I'm writing a web site app based on this PR. I think there is nothing like real-world usage to push a library through it's paces. So, this week, I'm doing that. There will be likely be odd small changes and tweaks showing up. :)

JohnAD:
Ack! I'm glad I'm using the database in a real-world probect. Lot of good cleanup.

Of those things, I'm getting:

```
/home/johnd/Projects/reposupport/src/views.nim(156, 52) template/generic instantiation of `async` from here
/home/johnd/Projects/norm/src/norm/mongodb.nim(282, 12) Warning: 'getOne' is not GC-safe as it accesses 'normDatabase' which is a global using GC'ed memory [GcUnsafe2]
/home/johnd/Projects/reposupport/src/views.nim(156, 52) template/generic instantiation of `async` from here
/home/johnd/Projects/reposupport/src/database.nim(89, 6) Warning: 'read_user' is not GC-safe as it calls 'getOne' [GcUnsafe2]
/home/johnd/Projects/reposupport/src/views.nim(156, 52) template/generic instantiation of `async` from here
```

When compiling for a threaded web project (async) that uses the `jester` library. I'm going to look into how to make the mongodb library thread-safe.

It's possible I might have to make a variant of the commands that do not rely on a global variable.

moigagoo:
@JohnAD Thanks for keeping the PR updated and for your contribution! Hope to see MongoDB support in Norm soon :-)

Would you mind sharing the code to your app that uses Norm with MongoDB? I'm like to peek into the code.

JohnAD:
It's early enough in the coding that I don't think my partners would mind. I'll send you an invite to the private repo to see it for a few days.

JohnAD:
BTW, as you will see in the code, I "broke apart" jester's routing to add more configuration in views.nim. But I think I'll be able to go back to the regular model soon now that I've resolved some of the async issues.

JohnAD:
I'm thinking of switching to the async version of `nimongo` and allowing an optional parameter somewhere for a timeout. Otherwise, if the server is partly hung, norm hangs forever.

Perhaps it could be an optional parameter on the operations. So, in addition to `myObject.insert()` to also allow `myObject.insert(timeout=4000)`; which responds with an error after 4 seconds of waiting.

This would be an easy change; though I don't believe there would be any good way to add a unit test for the timeout.

moigagoo:
@JohnAD FYI I'm going to merge a PR with some refactoring: https://github.com/moigagoo/norm/pull/25

Rowutils are now backend-specific, which allows to write backend-specific conversion rules.

JohnAD:
@moigagoo 

I have something of a conundrum.

My using norm in a real-world project, writing the web site, has been extremely useful and there are small but import changes for this PR coming soon.

However, in doing all this I've discovered something complicated: though the underlying `nimongo` library handles asynchronous communication from a batch point of view, it is not well suited to playing well nicely with other asynchronous libraries such as `Jester`. To make things work and scale properly, I ended up breaking many of the fundamentals of the library. In short, I ended up forking the library because my changes would have been very inappropriate to push into `nimongo`.

My new fork is here: https://github.com/JohnAD/mongopool  .

It actually runs very smoothly with Jester and I imagine it would the same with other threaded systems. So, for my own project, I'm sticking with my spinoff library.

So, for this PR, I have 4 options that I can think of:

1. I could switch the PR to using `mongopool` instead of `nimongo`. I've pretty much already done that in a branch.

2. I could stick with `nimongo` for the PR and keep my version of `norm` for `mongopool` private.

3. ~~I could write support for both as separate files.~~

4. I could write the mongodb support for norm to use compile-time parameters to choose either library ; using `when` statements to selectively choose either library and match the behavior to that library. The predictable way my code uses the library makes this quite possible.

Actually, even as I typed option #3 I realized just how much work that would be in the long run. I'd rather not do that.

But of options 1, 2, 4, or something else: any preference?

moigagoo:
@JohnAD Option 1 is the way to go:
- it's the shortest path to release
- it can be upgraded to option 4 in the future if necessary
- if your fork offers better compatibility than nimongo, I can't see any reasons to use nimongo

PurpleSquirrelJohn:
It has been a long time since you have seen any progress from me on my PR. In part because of me pausing to write the underlying `mongopool` db driver which required that I write an underlying `bson` spec library.

And, an odd thing happened along the way: I wanted the `bson` to be full and complete, so I had it encompass many of the same functions that the standard JSON library does. This included "marshalling" support where you can easily convert from BSON/JSON to and from an object.

At that point, I realized that most of the benefit of using `norm` disappeared as an unintended side effect. For example, using norm:

```
withDb:
  var p = Msg()
  pj.txt = "hello"
  p.x = 99 
  p.insert()
```

But without it, using marshalling:

```
var p = Msg()
pj.txt = "hello"
p.x = 99 
discard db.insertOne("msg", p.toBson)
```

Norm is still more concise, but not hugely so. In fact to test that, the next 4 web sites (that used nim) that I wrote specifically did not use Norm. And it worked really well.

I feel bad about abandoning the PR. But, I suspect that the effort to maintain the mongodb support function would not be worth it. 

I will continue to use Norm for SQL; it is great tool for that.

PR #25, opened 2019-08-21
Split backend-specific code.

moigagoo:


moigagoo:
@krux02 please provide more info.

krux02:
https://github.com/nim-lang/Nim/blob/d0e5bd2305db719b0c9acb6a017c8aa579a2f246/testament/important_packages.nim#L59

This package was tested as an important package for Nim CI. But the way this package was tested depended on the name of the Project. So I would not technically say it is your fault.

moigagoo:
@krux02 Is there anything I must do on my part to fix the tests? Should I notify anyone when I'm renaming files in the future?

krux02:
Well, in the link you see how your package is tested. I think you should discuss with @narimiran a transition phase. Ranaming a file is expected to exist in the test is not a valid transition period.

narimiran:
> Is there anything I must do on my part to fix the tests?

We discovered the problem and we renamed the file in our tests, so this is now fixed.

> Should I notify anyone when I'm renaming files in the future?

We only test `t(est)sqlite.nim`, so any other renaming should be fine.
But, since your package is at version >1.0, be aware of breaking stuff some other people might depend on.

PR #26, opened 2019-09-04
Allow splitting schema definition accross modules

moigagoo:
Currently, you describe your schema in type sections under db macro. This means, all your model types have to be defined in the same module, which may play poorly with your app's business logic.

This PR implements dbTypes and dbFromTypes macros that allow to mark any type section to be usable for schema generation and define a schema from existing types respectively.


JohnAD:
Excellent! When you are done with this PR, I will modify my PR to use the same naming.

moigagoo:
@JohnAD Please note that I had to do some significant refactoring in order to implement this feature without making a mess out of my code:
- SQL query generation has been moved into separate, backend-specific modules: sqlite/sqlgen and postgres/sqlgen
- Backend-agnostic proc to inject `id` field into type definition has been split into two and moved into a separate top-level module typedefutils.

Please do the same for Mongo, I think your code will benefit from it, too.

PR #33, opened 2019-10-11
Add default values to tables, deprecate default pragma

zedeus:
Didn't update the tests as it wasn't clear to me what they should look like, and I don't have PostgreSQL setup.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -8,11 +8,11 @@
 
 ## 1.0.18 (WIP)
 
+-   [!] Deprecate `default` pragma, always add default values to tables instead.
 -   [+] Add `createTable` and `dropTable`.
 -   [+] SQLite: Add means to write migrations: `addColumn`, `dropUnusedColumns`, `renameColumnFrom`, and `renameTableFrom`.
 -   [+] PostgreSQL: Add means to writemigrations: `addColumn`, `dropColumns`, `dropUnusedColumns`, `renameColumnFrom`, and `renameTableFrom`.
 -   [r] Rewrite table schema generation so that schemas are generated from typed nodes rather than untyped modes.
--   [r] Deprecate `default` pragma, always add default values to tables instead.
 
 
 ## 1.0.17 (September 12, 2019)
--- a/changelog.md
+++ b/changelog.md
@@ -12,6 +12,7 @@
 -   [+] SQLite: Add means to write migrations: `addColumn`, `dropUnusedColumns`, `renameColumnFrom`, and `renameTableFrom`.
 -   [+] PostgreSQL: Add means to writemigrations: `addColumn`, `dropColumns`, `dropUnusedColumns`, `renameColumnFrom`, and `renameTableFrom`.
 -   [r] Rewrite table schema generation so that schemas are generated from typed nodes rather than untyped modes.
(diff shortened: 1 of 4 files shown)


src/norm/pragmas.nim
```diff
@@ -30,8 +30,9 @@ template dbCol*(val: string) {.pragma.}
 template dbType*(val: string) {.pragma.}
   ## DB native type to use in table schema.
 
-template default*(val: string) {.pragma.}
+template default*(val: string) {.pragma, deprecated.}
   ## Default value for the DB column.
+  ## Deprecated since 1.0.18
```
moigagoo: Please keep the original docstring and add a notice to it instead of replacing it altogether. Also, please omit hardcoded version numbers since I'm not sure the next version will be 1.0.18 and not 1.1.0.

I suggest this:
```
##[ Default value for the DB column.

**Deprecated.** Default values are set automatically for all columns, this pragma is ignored.
]##
```

src/norm/pragmas.nim
```diff
@@ -30,8 +30,9 @@ template dbCol*(val: string) {.pragma.}
 template dbType*(val: string) {.pragma.}
   ## DB native type to use in table schema.
 
-template default*(val: string) {.pragma.}
+template default*(val: string) {.pragma, deprecated.}
```
moigagoo: Please add a deprecation notice:
```
{.pragma, deprecated: "Default values are set automatically, this is ignored.".}
```

src/norm/sqlite/sqlgen.nim
```diff
@@ -77,17 +77,17 @@ proc getDbType(fieldRepr: FieldRepr): string =
...
+      of "float": "REAL NOT NULL DEFAULT 0"
+      else: "TEXT NOT NULL DEFAULT ''"
     elif fieldRepr.typ.kind == nnkBracketExpr and $fieldRepr.typ[0] == "Option":
       case $fieldRepr.typ[1]
-        of "int", "Positive", "Natural", "bool", "DateTime": "INTEGER"
-        of "string": "TEXT"
-        of "float": "REAL"
-        else: "TEXT"
-    else: "TEXT NOT NULL"
+      of "int", "Positive", "Natural", "bool", "DateTime": "INTEGER "
```
moigagoo: Why the trailing space after INTEGER?

moigagoo: Also, why change the case formatting? I know it's debatable whether ofs should be indented but this is the way the rest of the case blocks are formatted, so I suggest we conform to it here.

changelog.md
```diff
@@ -12,6 +12,7 @@
 -   [+] SQLite: Add means to write migrations: `addColumn`, `dropUnusedColumns`, `renameColumnFrom`, and `renameTableFrom`.
 -   [+] PostgreSQL: Add means to writemigrations: `addColumn`, `dropColumns`, `dropUnusedColumns`, `renameColumnFrom`, and `renameTableFrom`.
 -   [r] Rewrite table schema generation so that schemas are generated from typed nodes rather than untyped modes.
+-   [r] Deprecate `default` pragma, always add default values to tables instead.
```
moigagoo: [r] → [!]

moigagoo:
@zedeus thanks!

PR #35, opened 2019-10-20
Fixes "unreachable statement" error in rowutils.

dom96:


moigagoo:
@dom96 

1. Is Postgres backend affected too? If yes, please submit a fix for it as well.
2. Isn't there a better way, like rewriting the condition? I'm OK with the hack but if there's a more idiomatic fix, I'd rather pick it.
3. Please update the changelog in this PR as well.

dom96:
> Is Postgres backend affected too? If yes, please submit a fix for it as well.

I looked at the code and didn't see anything similar, so I don't think so.

> Isn't there a better way, like rewriting the condition? I'm OK with the hack but if there's a more idiomatic fix, I'd rather pick it.

Not that I know of.

> Please update the changelog in this PR as well.

Done.

PR #43, opened 2019-10-31
Rename pragma `table` to `dbTable`

### LINKED ISSUE #39: Rename table to dbTable
Pragma `table` should be called `dbTable` to keep in line with `dbCol` and `dbType`.

sambercx:
Fixes #39

### DIFF
--- a/src/norm/pragmas.nim
+++ b/src/norm/pragmas.nim
@@ -49,5 +49,8 @@ template onDelete*(val: string) {.pragma.}
 template onUpdate*(val: string) {.pragma.}
   ## Add an ``ON UPDATE <POLICY>`` constraint.
 
+template table*(val: string) {.pragma, deprecated: "`table` has been renamed to `dbTable`".}
+  ## Set table name. Lowercased type name is used when unset.
+
 template dbTable*(val: string) {.pragma.}
   ## Set table name. Lowercased type name is used when unset.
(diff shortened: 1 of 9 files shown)


src/norm/pragmas.nim
```diff
@@ -49,5 +49,5 @@ template onDelete*(val: string) {.pragma.}
 template onUpdate*(val: string) {.pragma.}
   ## Add an ``ON UPDATE <POLICY>`` constraint.
 
-template table*(val: string) {.pragma.}
+template dbTable*(val: string) {.pragma.}
```
moigagoo: Please keep the old pragma bu mark it as deprecated. It's better to give Norm users some time to update their code.

src/norm/sqlite/sqlgen.nim
```diff
@@ -23,15 +23,15 @@ proc getTable*(objRepr: ObjRepr): string =
   result = objRepr.signature.name.toLowerAscii()
 
   for prag in objRepr.signature.pragmas:
-    if prag.name == "table" and prag.kind == pkKval:
+    if prag.name == "dbTable" and prag.kind == pkKval:
```
moigagoo: Please check for both dbTable and table for the transition period. We'll remove it after a couple of minor versions. A TODO mark is fine.

sambercx:
Done!

PR #44, opened 2019-11-01
Changed getMany's limit parameter to an optional parameter.

Dankr4d:
I've set the getMany's limit parameter to an optional parameter with default value -1. On sqlite we can pass a negtaive number to ignore `LIMIT`. On postgresql negative numbers are not allowed (instead of a negative number it must be `LIMIT ALL`). So I've modified the `genGetManyQuery`, that `LIMIT` will not be added to the query if the parameter `withLimit` (default true) is set to false. I think it's not the optimal solution, but I have not found a better for this.

### DIFF
--- a/src/norm/postgres.nim
+++ b/src/norm/postgres.nim
@@ -222,7 +222,7 @@ template genWithDb(connection, user, password, database: string, dbTypeNames: op
 
         result.getOne(id)
 
-      proc getMany(objs: var seq[object], limit: int, offset = 0,
+      proc getMany(objs: var seq[object], limit: int = -1, offset = 0,
                    cond = "TRUE", params: varargs[DbValue, dbValue]) {.used.} =
         ##[ Read ``limit`` records with ``offset`` from DB into an existing open array of objects.
 
@@ -232,16 +232,21 @@ template genWithDb(connection, user, password, database: string, dbTypeNames: op
         if len(objs) == 0: return
 
         let
-          getManyQuery = genGetManyQuery(objs[0], cond, len(params))
-          params = @params & @[?min(limit, len(objs)), ?offset]
+          withLimit = if limit < 0: false else: true
+          getManyQuery = genGetManyQuery(objs[0], cond, len(params), withLimit = withLimit)
+        var params = @params
+        if withLimit:
+          params.add(?min(limit, len(objs)))
+        params.add(?offset)
 
         debug getManyQuery, " <- ", params.join(", ")
-
         let rows = dbConn.getAllRows(getManyQuery, params)
+
+        if limit < 0: objs.setLen len(rows)
 
         rows.to(objs)
 
-      proc getMany(T: typedesc, limit: int, offset = 0,
+      proc getMany(T: typedesc, limit: int = -1, offset = 0,
                    cond = "TRUE", params: varargs[DbValue, dbValue]): seq[T] {.used.} =
         ##[ Read ``limit`` records  with ``offset`` from DB into a sequence of objects,
         create the sequence on the fly.
@@ -249,7 +254,9 @@ template genWithDb(connection, user, password, database: string, dbTypeNames: op
         Filter using ``cond`` condition.
         ]##
 
-        result.setLen limit
+        # Setting result len to 1 if limit is lower than 0 because genGetManyQuery
+        # requieres at least one object to get the db table name
+        result.setLen if limit < 0: 1 else: limit
         result.getMany(limit, offset, cond, params)
 
       template update(obj: object, force = false) {.used.} =
--- a/src/norm/postgres.nim
+++ b/src/norm/postgres.nim
@@ -222,7 +222,7 @@ template genWithDb(connection, user, password, database: string, dbTypeNames: op
 
         result.getOne(id)
 
-      proc getMany(objs: var seq[object], limit: int = -1, offset = 0,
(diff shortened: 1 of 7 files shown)


moigagoo:
Thanks for submitting a PR!

I think I can see a cleaner solution:
1. Add procs `getAll` to sqlite and postgres, which are just `getMany` but without `limit` or `offset`.
2. Add `genGetAllQuery` to sqlite/sqlgen and postgres/sqlgen, which are just `genGetManyQuery` but wuthout LIMIT or OFFSET.

A slightly different flavor of this solution is to add `getMany` procs without `limit` or `offset` instead of `getAll`.

What do you think? I'm leaning towards `getAll` because getting all rows without limits is generally speaking a dangerous operation since you don't control the amount of data passed.

Dankr4d:
You're absolutely right!
I also favor the `getAll` variant as you mentioned. That's a much cleaner solution. Thank you for reply. I'll get to work.

src/norm/sqlite.nim
```diff
@@ -252,6 +252,33 @@ template genWithDb(connection, user, password, database: string, dbTypeNames: op
...
+        rows.to(objs)
+
+      proc getAll(T: typedesc, cond = "1", params: varargs[DbValue, dbValue]): seq[T] {.used.} =
+        ##[ Read all records from DB into a sequence of objects, create the sequence on the fly.
+        This is a dangerous operation as you don't control the amount of data received.
+
+        Filter using ``cond`` condition.
+        ]##
+
+        result.setLen 1
```
moigagoo: Is this necessary? Container length is set to rows.len in line 268 anyway.

src/norm/postgres.nim
```diff
@@ -252,6 +252,33 @@ template genWithDb(connection, user, password, database: string, dbTypeNames: op
...
+        rows.to(objs)
+
+      proc getAll(T: typedesc, cond = "TRUE", params: varargs[DbValue, dbValue]): seq[T] {.used.} =
+        ##[ Read all records from DB into a sequence of objects, create the sequence on the fly.
+        This is a dangerous operation as you don't control the amount of data received.
+
+        Filter using ``cond`` condition.
+        ]##
+
+        result.setLen 1
```
moigagoo: Is this necessary? Container length is set to rows.len in line 268 anyway.

Dankr4d:
Here I am.
I tried to implement the getAll procs that I would only pass the model type to the `genGetAllQuery` proc. But I'm not familiar with macros and templates (and then I was busy).
In your `genGetManyQuery` procs you get the table name from the `typedesc` and the table columns from the object. I also saw, that you added a `getColumns` macro with the `typdesc` as first param. But i didn't get it work. The solution how it works now, is setting the length from the seq in the getAll proc (operating on the `typdesc`) to 1, that the next `getAll` proc (operating on a seq of objects) can read out the object for the `genGetAllQuery` proc. I'm not so happy with this solution, because after querying the rows the seq length will reset again. If no row is received, the first `setLen` (to 1) is unnecessary.

moigagoo:
Oh, now I see, there's this objs[0] bit that requires objs to be non-empty. It seems like having a flavor of getAll that accepts a sequence is useless: you want to get all rows by definition, regardless of the container size.

I've cloned your fork and tried making getAll work with typedesc, to no success so far but I think I'm getting there. I'll send a PR if I have anything interesting to share.

Could you please add some tests anyway?

Dankr4d:
Thank you for your help!
> Could you please add some tests anyway?

Yes, I'm already on it.

moigagoo:
> Thank you for your help!

No, thank _you_ for your help. Getting contributions means a lot for me as a package maintainer, and I'd just like to express my appreciation for you taking the time to submit the PR once again.

I got the getAll problem solved, will send you a PR. Today or tomorrow probably, when I get the time to code it in.

moigagoo:
@Dankr4d here you go https://gist.github.com/moigagoo/23d48dc277192b5b6559c3d8ab0f9190

Dankr4d:
I wrote some tests. But I couldn't test them with postgresql.
```bash
[dankrad@TNB ~]$ postgres --version
postgres (PostgreSQL) 11.5
```
I've received the following error:
```nim
[Suite] Creating and dropping tables, CRUD
NOTICE:  table "users" does not exist, skipping
NOTICE:  table "publishers" does not exist, skipping
NOTICE:  table "books" does not exist, skipping
NOTICE:  table "editions" does not exist, skipping
Traceback (most recent call last)
/home/dankrad/projects/norm/src/norm/postgres.nim(173) tpostgres
/home/dankrad/projects/norm/src/norm/postgres/sqlgen.nim(216) genInsertQuery
/opt/Nim/lib/pure/strutils.nim(1775) join
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
Error: execution of an external program failed: '/home/dankrad/projects/norm/tests/tpostgres '
```
Should I create an issue? 

But anyway these are the same tests as in `tsqlite.nim` so the CI should pass.

moigagoo:
Are you running the tests in Docker? Also, try running them with my patch. I've tested it, the tests pass (at least my very basic ones included in the patch).

Dankr4d:
I've applied your patch and tested them with docker and native. All tests passed through successfully. Due the merge of your patch I've removed my "Create records" tests. Should I re add them? I've also removed the "Read records" tests since the proc for reading them into a sequence has been removed.

PR #52, opened 2020-01-14
Attempt to deprecate NOT NULL

3n-k1:
I attempted to replicate [this pull](https://github.com/moigagoo/norm/pull/33/files) for `NOT NULL`. I'm not sure if anything else is needed, but it looks good to me.

### DIFF
--- a/src/norm/pragmas.nim
+++ b/src/norm/pragmas.nim
@@ -34,8 +34,12 @@ template default*(val: string) {.
   pragma, deprecated: "Default values are set automatically, this is ignored.".}
   ## Default value for the DB column.
 
-template notNull* {.pragma.}
-  ## Add ``NOT NULL`` constraint.
+template notNull* {.
+  pragma, deprecated: "NOT NULL is the default with all fields exception Options, this is ignored.".}
+  ##[ Add ``NOT NULL`` constraint.
+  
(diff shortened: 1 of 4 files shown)


src/norm/pragmas.nim
```diff
@@ -34,8 +34,12 @@ template default*(val: string) {.
   pragma, deprecated: "Default values are set automatically, this is ignored.".}
   ## Default value for the DB column.
 
-template notNull* {.pragma.}
-  ## Add ``NOT NULL`` constraint.
+template notNull* {.
+  pragma, deprecated: "NOT NULL is the default with all fields exception Options, this is ignored.".}
+  ##[ Add ``NOT NULL`` constraint.
+  
+  **Deprecated.** NOT NULL is the default with all fields exception Options, this is ignored.
```
moigagoo: This is unnecessary since you've already added the deprecation message to `deprecated` pragma.

3n-k1:
@moigagoo What do I need to do to port this to sqlite/sqlgen.nim? Looking through the file, all I can think is to remove [this line](https://github.com/moigagoo/norm/blob/develop/src/norm/sqlite/sqlgen.nim#L105) and move `result.add " NOT NULL"` outside the for loop. Would that be sufficient?

3n-k1:
Alright, changes made. Let me know if they look good and I'll squash everything.

PR #53, opened 2020-01-14
Enable automatic foreign keys

### LINKED ISSUE #40: Automatic foreign key handling
Currently, Norm models are deliberately very close to the generated table schema. To define a one-to-many relation, you just mark a field with `fk` pragma and reference another model in it.

However, key benefit of having an ORM is that it gives you a nicer interface to your data, so you don't want a field that contains an id of the referenced object like in a DB, you want the referenced object itself. Currently, to achieve that, you should write a parser manually (see https://github.com/moigagoo/norm-sample-webapp/blob/develop/src/models.nim#L19)

To make Norm a more useful ORM, we should handle foreign keys automatically. This:
```
type
  User = object
    name: string
  Pet = object
    owner: User
```
should be equivalent to this:
```
type
  User = object
    name: string
 Pet = object
    owner {.
      fk: User,
      dbCol: "ownerId",
      dbType: "INTEGER NOT NULL",
      parseIt: User.getOne(it.i.int),
      formatIt: ?it.id
    .}: User
```

3n-k1:
Enable automatic detection of foreign keys, and automatically add the correct pragmas.
This currently only works when a model and its fk are in the same `type` block. Closes #40.

Marked as WIP because the code needs a little bit of fixing up. I'm currently just submitting this PR for review. Once everything looks good, I'll clean up the code and make things look nice. I'll also need to write tests, but so far it looks like sqlite works fine.


3n-k1:
Also, tests will probably need to be written

3n-k1:
I dug around in the code a bit and saw the `dbFromTypes` macro. If this code looks good, I'll work on altering that macro so that models in different modules can have foreign keys on each other.

3n-k1:
after some experimentation, I don't think it'll be possible to automatically add foreign keys for `dbFromTypes`.

I'll add some tests and then remove the WIP tag.

3n-k1:
Issue with the postgres test. Something in the nim compiler is erroring at `lib/core/macros.nim`, line 1345. I think docker is just running a version of nim that's too old, but I might be wrong. Either way, the sqlite test runs perfectly.

moigagoo:
Hi! Sorry for the slow response. I forgot to subscribe to the PR and missed all you the latest updates.

The PR is good, there are just two minor conflicts and it's good to merge.

However, I'd like to test it for some corner cases and maybe try an alternative approach first. I have a feeling there's a different way to implement it but with universal db and dbFromTypes support. I may be wrong though, haven't dug into it too deep yet.

3n-k1:
Hey, no problem, I totally understand :)

I'll go ahead and fix the conflicts today, let me know when you've dug into things and what you find.

Issue #87, opened 2020-10-22
Unique and Foreign Key pragma ?


Clonkk:
Hello,

I don't see any foreign key or unique pragma for Model definition.

I know that for Foreign key are handled by using a Model as a field.

```nim
type
  A = ref object of Model
    field : string # stirng of size 1024*1024*1024

  B = ref object of Model
    a: A
```

But this approach can have some issues (unless I'm mistaken): 
* You have to load the full object in memory. For large objects (that can also contains foreign key) it's a problem.
* You can't create foreign key on anything else than the id field.

I think there are valid use case for wanting this :   
```nim
type
  A = ref object of Model
    field : string 
    name {. unique .}: string

  B = ref object of Model
    a {. fk: A.id .}: int
```

Or this : 
```nim
type
  A = ref object of Model
    name {. unique .}: string

  B = ref object of Model
    a {. fk: A.name .}: int
```

moigagoo:
This is the way it used to be done in Norm 1. You had to create those foreign keys manually, and it was a PITA to instantiate objects with other objects as fields.

So I've taken a more opinionated approach with Norm 2 and made it less flexible but more predictable and reliable.

Adding unique pragma should be trivial. I don't have the time now and I'm focusing on the issue with fetching NULL values. So feel free to send a PR.

I also think it is possible to add customizable foreign keys but this is something I probably won't have the time to invest into any time soon.

Clonkk:
> You had to create those foreign keys manually, and it was a PITA to instantiate objects with other objects as fields.

Yes, I agree using manually declared foreign key is more work than your current solution; but I can't see why we can't have both.
If you declare manually a foreign key, then selecting a table will only return the foreign key values and fetching the datas should be the user's responsability.  

> So feel free to send a PR.

I'll work on a proposal.

moigagoo:
I have a couple of ideas regarding memory usage optimization. Like introducing lazy mode that would not automatically fetch related objects, or allowing Model fields to be nil. It's just that I don't have the opportunity to do all that unfortunately. Again, feel free to contribute.

PR #89, opened 2020-10-23
Feat pragmas unique fk

Clonkk:
PR for https://github.com/moigagoo/norm/issues/87

* Add unique pragma
* Add a fk pragma to use foreign key by hand. Foreign key reference the id field of a table. 
  * Add PRAGMA foreign_keys = ON to SQLite when using foreign keys
* Hide debug behind a compilte time define because : 
  * it was slowing down the application (even with no logging activated)
  * In case of large binary data it was making debug log unusable

### DIFF
--- a/src/norm/postgres.nim
+++ b/src/norm/postgres.nim
@@ -98,6 +98,13 @@ proc createTables*[T: Model](dbConn; obj: T) =
       fkGroups.add "FOREIGN KEY($#) REFERENCES $#($#)" %
         [obj.col(fld), typeof(get val.model).table, typeof(get val.model).col("id")]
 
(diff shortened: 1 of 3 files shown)


Clonkk:
If you agree with the spirit of the change, i'll go ahead and add entry to CHANGELOG, README and some tests

moigagoo:
@Clonkk could you please split that into three PRs? One for compile time logging define, one for unique pragma, and one for custom fks. The first one is trivial and can be merged immediately. The second one is not trivial but still is pretty simple, I can review it quickly. The lat one however, is tricky, I need to take a very good look at it.

PR #91, opened 2020-10-24
Added unique pragma

Clonkk:
https://github.com/moigagoo/norm/issues/87
* Added unique pragma


src/norm/pragmas.nim
```diff
@@ -16,3 +16,5 @@ template ro* {.pragma.}
 
   ``id`` field is ``ro`` by default.
   ]##
+
+template unique* {.pragma.}
```
moigagoo, line 20: @Clonkk this needs a docstring.

moigagoo (changes requested):
@Clonkk this looks pretty obvious but it's hard to see how that pragma affects the way queries are generated and what's the profit of using it for the user. This should be addressed in tests.

Also, please add a changelog entry for this change.

moigagoo (changes requested):
Please add a test and resolve the conflict in the changelog.

changelog.rst
```diff
@@ -96,8 +96,11 @@ The new algorithm adds alias for each joined table. The alias is named after the
 
 -   [r] Update Nim version to 1.4.0 in Dockerfile.
 
+-   [+] Added a define to enable logging
```
moigagoo, line 99: This line is redundant, see the line below.

src/norm/pragmas.nim
```diff
@@ -16,3 +16,7 @@ template ro* {.pragma.}
...
+##[ Mark field  as unique
```
moigagoo, line 21: This should be a single line comment.

tests/tpostgresunique.nim
```diff
@@ -0,0 +1,55 @@
...
+import models
+
+
+const
+  dbHost = "postgres"
+  dbUser = "postgres"
+  dbPassword = "postgres"
+  dbDatabase = "postgres"
+
+proc duplicateInsert(dbConn: DbConn)
```
moigagoo, line 15: What's the purpose of this forward declaration?

tests/tpostgresunique.nim
```diff
@@ -0,0 +1,55 @@
...
+  teardown:
+    close dbConn
+    resetDb()
+
+  test "Duplicate insert":
+    # Check test fail with DbError
+    expect DbError:
+      duplicateInsert(dbConn)
+
+proc duplicateInsert(dbConn: DbConn)=
```
moigagoo, line 37: I think this is way overcomplicated. You could just have all that as part of the test, inside the test body.

tests/tpostgresunique.nim
```diff
@@ -0,0 +1,55 @@
...
+  block:
+    let
+      toy = newToy(123.45)
+      pet = newPet("cat", toy)
+
+    var person = newPerson("Alice", pet)
+    dbConn.insert(person)
+    check true
```
moigagoo, line 45: I'm not really happy with those check true and check false things. Explicit, meaningful checks are better.

tests/tpostgresunique.nim
```diff
@@ -0,0 +1,55 @@
...
+
+  block:
+    let
+      toy = newToy(36.66)
+      pet = newPet("dog", toy)
+
+    var person = newPerson("Alice", pet)
+    dbConn.insert(person)
+    # Check that test does not reach this point
```
moigagoo, line 54: If the whole proc were just part of the test, there would be no need in this trick.

tests/tpostgresunique.nim
```diff
@@ -0,0 +1,55 @@
...
+  setup:
+    resetDb()
+    let dbConn = open(dbHost, dbUser, dbPassword, dbDatabase)
+
+  teardown:
+    close dbConn
+    resetDb()
+
+  test "Duplicate insert":
+    # Check test fail with DbError
```
moigagoo, line 33: Unnecessary comment, it's pretty obvious what the check is from the next line.

tests/tsqliteunique.nim
```diff
@@ -0,0 +1,52 @@
...
+import options
+import os
+
+import norm/[model, sqlite]
+
+import models
+
+const dbFile = "test.db"
+
+proc duplicateInsert(dbConn: DbConn)
```
moigagoo, line 12: See comments for Postgres.

moigagoo (changes requested):

Clonkk:
> What's the point of this forward declaration?

It's mostly habits from developing in C.

moigagoo (approved):
I'm still not quite happy with some details, but it'll be faster for me to just fix them myself after merge.

PR #92, opened 2020-10-24
Hide debug qry behind compile time define

Clonkk:
* debug qry is now behind a compile time define

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -12,8 +12,6 @@ Changelog
 2.2.0 (WIP)
 ===========
 
--   [+] Added a define to enable logging
-
 -   [!][f][t] The way ``JOIN`` statements are generated has been changed competely. The previous algorithm was just wrong, it didn't work with models that that multiple FKs to the same model or when the same model was referenced from the root model and any of its ``Model`` fields or their ``Model`` fields.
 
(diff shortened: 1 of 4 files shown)


src/norm/sqlite.nim
```diff
@@ -95,7 +95,9 @@ proc createTables*[T: Model](dbConn; obj: T) =
 
   let qry = "CREATE TABLE IF NOT EXISTS $#($#)" % [T.table, (colGroups & fkGroups).join(", ")]
 
-  debug qry
+  when defined(normDebug):
+    debug qry
+
```
moigagoo, line 100: @Clonkk you haven't put an blank line between debug and exec in other cases, so, for consistency, I'm asking you to remove it here.

moigagoo (changes requested):
Thanks for the PR! Looks fine to me, apart from a very minor blank line thing.

Please add a new changelog entry and update the readme as this change changes the way debugging works, this needs to be addressed in the tutorial.

changelog.rst
```diff
@@ -12,6 +12,8 @@ Changelog
 2.2.0 (WIP)
 ===========
 
+-   [+] Added a define to enable logging
```
moigagoo, line 15: Please move that to the bottom of the 2.2.0 section. I want the breaking changes to stand out.

moigagoo (changes requested):

Issue #84, opened 2020-09-18
Switch from nimble test to testament


moigagoo:
https://nim-lang.github.io/Nim/testament.html

Clonkk:
I have a branch where both nimble and testament pass all tests. It also cleans up the tests added for foreign keys.
 
Here's the PR : https://github.com/moigagoo/norm/pull/94

I haven't modified travis to run with testament so you can make the change whenever you want.

moigagoo:
Moving to testament of something I'm not sure of. Also, the statement in the title is just wrong: it's not unittest vs testament, it's nimble's test runner vs testament. I didn't research the topic well enough.

So, what I'm doing now is removing testament comments from tests to keep things clear.

I like when one PR fixes one problem, it's easier to review that way. Refactoring tests is a separate task, so I'm removing its implementation from develop _for now_.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

PR #93, opened 2020-10-24
Added fk pragma 

Clonkk:
https://github.com/moigagoo/norm/issues/87

* Added fk pragmas on integer field that will create FK referencing the id field of the table.
* Added check that the pragma target is a model
The idea is that it should be simple; Pragma is applied on an integer field and target a Model.


src/norm/postgres.nim
```diff
@@ -98,6 +98,14 @@ proc createTables*[T: Model](dbConn; obj: T) =
       fkGroups.add "FOREIGN KEY($#) REFERENCES $#($#)" %
         [obj.col(fld), typeof(get val.model).table, typeof(get val.model).col("id")]
 
+    when obj.dot(fld).hasCustomPragma(fk):
+      const pragVal = $(obj.dot(fld).getCustomPragmaVal(fk))
```
moigagoo, line 102: Could be one `const` block.

moigagoo: I'm not comfortable with that string conversion. What is the point of that line? If you need the table name, use `table`.

src/norm/postgres.nim
```diff
@@ -98,6 +98,14 @@ proc createTables*[T: Model](dbConn; obj: T) =
...
+      const fkTargetIsModel = isModel(none(typeof(obj.dot(fld).getCustomPragmaVal(fk))))
```
moigagoo, line 103: Why are you constructing a `None` and checking that against `is Model` instead of just checking the type? `isModel` is for object checks, not type checks. In other words, `obj.dot(fld).getCustomPragmaVal(fk) is Model` would probably do the trick.

src/norm/postgres.nim
```diff
@@ -98,6 +98,14 @@ proc createTables*[T: Model](dbConn; obj: T) =
...
+      when val is SomeInteger and fkTargetIsModel:
+        fkGroups.add "FOREIGN KEY ($#) REFERENCES $#(id)" % [fld, $pragVal]
+      else:
+        {.error: "Pragma fk must be used on SomeInteger field and target a Model." .}
```
moigagoo, line 107: Shouldn't that be a `raise newException`? Never seen this syntax before.

src/norm/pragmas.nim
```diff
@@ -16,3 +16,8 @@ template ro* {.pragma.}
 
   ``id`` field is ``ro`` by default.
   ]##
+
+template fk*(val: typed) {.pragma.}
+  ##[ Mark field as foreign key another type. ``val`` is either a type or a "type.field"
+  expression. If a type is provided, its ``id`` field is referenced.
```
moigagoo, line 22: This seems to be copypasted from Norm 1.0. AFAIU `fk` value can only be a Model type, and the whole thing works based on the assumption that `fld` is integer and thus can be used as a foreign key to the referenced model `id` field, which is integer too.

src/norm/sqlite.nim
```diff
@@ -91,10 +91,23 @@ proc createTables*[T: Model](dbConn; obj: T) =
...
+      when val is SomeInteger and fkTargetIsModel:
+        fkGroups.add "FOREIGN KEY ($#) REFERENCES $#(id)" % [fld, $pragVal]
+      else:
+        {.error: "Pragma fk must be used on SomeInteger field and target a Model." .}
+
     colGroups.add colShmParts.join(" ")
 
   let qry = "CREATE TABLE IF NOT EXISTS $#($#)" % [T.table, (colGroups & fkGroups).join(", ")]
 
+  # Enable foreign key
```
moigagoo, line 106: I'm not sure this is necessary. Foreign keys are working just fine with SQLite right now, without this explicit pragma.

moigagoo (changes requested):

moigagoo:
@Clonkk Thanks for the PR!

Could you please provide some reasoning behind this proposal? You see, a lot of logic in the current implementation is based on the fact `Model` fields in Nim are turned into foreign keys in SQL, and that this mapping is based on the `id` field. See the code for `select` that generates those complex JOINs to better understand what I'm talking about.

I'm not sure the changes proposed will work with that. To prove me wrong, please try running some tests with `fk` pragma. Specifically, try doing some `select`ing.

src/norm/postgres.nim
Clonkk: Yeah,  that's a better way to handle it

src/norm/postgres.nim
Clonkk: Do you mean the error pragma ? An error output by the compiler seemd more approriate. I'll improve the message.

src/norm/postgres.nim
Clonkk: table is better than string conversion indeed.

src/norm/sqlite.nim
Clonkk: I agree that with or without, I haven't noticed any change but the official doc of sqlite3 says it's necessary to enable it by hand : 

https://sqlite.org/pragma.html#pragma_foreign_keys 
 > As of SQLite version 3.6.19, the default setting for foreign key enforcement is OFF. However, that might change in a future release of SQLite. The default setting for foreign key enforcement can be specified at compile-time using the SQLITE_DEFAULT_FOREIGN_KEYS preprocessor macro. To minimize future problems, applications should set the foreign key enforcement flag as required by the application and not depend on the default setting. 

https://sqlite.org/foreignkeys.html (chapter 2)
 > Assuming the library is compiled with foreign key constraints enabled, it must still be enabled by the application at runtime, using the PRAGMA foreign_keys command

It could still be done by the user at the application level ?

Clonkk:
> Could you please provide some reasoning behind this proposal? You see, a lot of logic in the current implementation is based on the fact Model fields in Nim are turned into foreign keys in SQL, and that this mapping is based on the id field. See the code for select that generates those complex JOINs to better understand what I'm talking about.

The idea behind the pragma is to add a manual mode to foreign key handling. I kept the constraint that foreign keys must reference the field ``id`` of a Model. But it adds the possibility of declaring a foreign key explicitly by pragma only on an integer. 

I'm not sure if it should restricted to ``int`` only of if ``SomeInteger`` is acceptable ?

Declaring manually a foreign key implies that it is handled by the developer. 
Unless I'm mistaken, the JOIN is done on field of type ``Model`` while the manual foreign key is an integer.
A select query will select the stored foreign key value (and not the fields belonging to the table of the foreign key). 
From what I've seen, you can still use both method in the same table: from norm perspective and for the select request, the manual foreign key is an integer.  

Example : 
```nim

import macros
import logging
import strutils
import options

import norm/model
import norm/pragmas
import norm/sqlite

var consoleLog = newConsoleLogger()
addHandler(consoleLog)

type
  Fooz = object
    s: int

  Foo = ref object of Model
    a : int
    b: float

  Baz = ref object of Model
    value: float64

  Bar = ref object of Model
    aaa {. fk: Foo .}: int
    # Does not compile : Fooz is not a Model
    # aaa {. fk: Fooz .}: int
    # Does not compile : aaa is not an int
    # aaa {. fk: Foo .}: float
    # Can check both an already existinf foreign keys and a new field
    bbb : Foo
    ccc: Baz

proc newFoo(): Foo=
  Foo(a: 0, b: 0.0)

proc newBaz(): Baz=
  Baz(value: 36.366)

proc newBar(): Bar=
  # Bar(aaa: 0, bbb: "123".DbBlob, bbb: newBaz())
  Bar(aaa: 0, bbb: newFoo(), ccc: newBaz())

let db = open("./local_sqlite.db", "", "", "")

proc createTable()=
  db.exec(sql"DROP TABLE IF EXISTS Foo")
  db.exec(sql"DROP TABLE IF EXISTS Bar")
  db.exec(sql"DROP TABLE IF EXISTS Baz")
  # Tables created by foreign keys
  # db.createTables(newFoo())
  # db.createTables(newBaz())
  db.createTables(newBar())


proc insertMe()=
  var foo : Foo = Foo(a: 11, b: 12.36)
  db.insert(foo)

  var bar : Bar = Bar(aaa: foo.id, bbb: foo, ccc:Baz(value: 63.33))
  db.insert(bar)

proc selectMe()=
  var foo = newFoo()
  db.select(foo, "Foo.a = ?", 11)

  var bar = newBar()
  db.select(bar, "Bar.aaa = ?", foo.id)
  echo "bar.id => ", bar[].id, " , bar.aaa => ", bar[].aaa, " => ", bar.bbb[], " => ", bar.ccc.value

createTable()
insertMe()
selectMe()
db.close()
```
Output : 
```
DEBUG CREATE TABLE IF NOT EXISTS "Foo"(a INTEGER NOT NULL, b FLOAT NOT NULL, id INTEGER NOT NULL PRIMARY KEY)
DEBUG CREATE TABLE IF NOT EXISTS "Baz"(value FLOAT NOT NULL, id INTEGER NOT NULL PRIMARY KEY)
DEBUG CREATE TABLE IF NOT EXISTS "Bar"(aaa INTEGER NOT NULL, bbb INTEGER NOT NULL, ccc INTEGER NOT NULL, id INTEGER NOT NULL PRIMARY KEY, FOREIGN KEY (aaa) REFERENCES "Foo"(id), FOREIGN KEY(bbb) REFERENCES "Foo"(id), FOREIGN KEY(ccc) REFERENCES "Baz"(id))
DEBUG INSERT INTO "Foo" (a, b) VALUES(?, ?) <- @[11, 12.36]
DEBUG INSERT INTO "Baz" (value) VALUES(?) <- @[63.33]
DEBUG INSERT INTO "Bar" (aaa, bbb, ccc) VALUES(?, ?, ?) <- @[1, 1, 1]
DEBUG SELECT "Foo".a, "Foo".b, "Foo".id FROM "Foo"  WHERE Foo.a = ? <- [11]
DEBUG SELECT "Bar".aaa, "Bar".bbb, "bbb".a, "bbb".b, "bbb".id, "Bar".ccc, "ccc".value, "ccc".id, "Bar".id FROM "Bar" LEFT JOIN "Foo" AS "bbb" ON "Bar".bbb = "bbb".id LEFT JOIN "Baz" AS "ccc" ON "Bar".ccc = "ccc".id WHERE Bar.aaa = ? <- [1]
bar.id => 1 , bar.aaa => 1 => (a: 11, b: 12.36, id: 1) => 63.33
```

src/norm/pragmas.nim
Clonkk: Fixed it.

Clonkk:
I can add some tests / changelog / readme entry once we've reached an agreement on design.

src/norm/postgres.nim
moigagoo: @Clonkk Oh, is that the way to emit compile time errors? I thought you should just raise an exception in those cases, compile and runtime alike.

src/norm/sqlite.nim
moigagoo: > It could still be done by the user at the application level ?

If you absolutely must, I think you can just add `exec sql "PRAGMA foreign_keys = ON"` into you app ¯\_(ツ)_/¯

Clonkk: Okay, I'm fine with that :smiley:

src/norm/postgres.nim
Clonkk: Yes, it is. You can also use {.fatal: "..." .}. 
I think raising a compiler error is cleaner than an exception in this case considering it's a compile constraint.

Clonkk:
I added tests for Foreign Keys.
The tests that are supposed to fail are skipped for now but run with testament when https://github.com/moigagoo/norm/issues/84 will be started.

What do you think otherwise about this way of handling manually a foreign keys ? 

If possible, (I don't want to sound entitled or to rush you :) ), It would be convenient for me  to have this topic resolved soon (I don't need a release, I can work on a devel version) as it's the only thing blocking for my migration from norm 1.1 -> 2.2.

Once it's resolved, I'll try to work on https://github.com/moigagoo/norm/issues/84 (since it would improve validation of foreign key handling) if that's okay with you.

moigagoo:
> What do you think otherwise about this way of handling manually a foreign keys ?

Before I answer that, I want to take a moment to thank you for your contributions to Norm. Knowing that somewhere, someone actually uses Norm fills my heart with joy. Thank you!

Now, to the question. I think I now understand the reasoning behind your PR, after reading the README addition you submitted. The point old to omit Norm's automatic foreign keys resolution as it can lead to unpotimal object allocations. I'm AFK now, but I'll try to review and merge you PR as soon as I can. Probably, this week.

Clonkk:
Hey man, Norm is awesome ! 
I hate writing SQL 😄

src/norm/postgres.nim
```diff
@@ -101,6 +101,20 @@ proc createTables*[T: Model](dbConn; obj: T) =
       fkGroups.add "FOREIGN KEY($#) REFERENCES $#($#)" %
         [obj.col(fld), typeof(get val.model).table, typeof(get val.model).col("id")]
 
+    when obj.dot(fld).hasCustomPragma(fk):
+      # Check val is int
+      when val is not int:
+        {.error: "Pragma fk must be used on an integer field. " & fld & " is not an integer." .}
```
moigagoo, line 107: @Clonkk I've read the manual on those pragmas. They generate compile-time messages but the compilation goes through. Is that the desired behavior? I mean, won't it it be better to crash on invalid model? It's an honest question, I don't think there is an obvious right answer.

src/norm/postgres.nim
```diff
@@ -101,6 +101,20 @@ proc createTables*[T: Model](dbConn; obj: T) =
       fkGroups.add "FOREIGN KEY($#) REFERENCES $#($#)" %
         [obj.col(fld), typeof(get val.model).table, typeof(get val.model).col("id")]
 
+    when obj.dot(fld).hasCustomPragma(fk):
+      # Check val is int
+      when val is not int:
```
moigagoo, line 106: I'm not sure this is idiomatic. I mean, shouldn't that be `val isNot int` or `not val is int`? Again, honest question.

src/norm/postgres.nim
```diff
@@ -101,6 +101,20 @@ proc createTables*[T: Model](dbConn; obj: T) =
...
+        {.error: "Pragma fk must be used on an integer field. " & fld & " is not an integer." .}
+      else:
+        # Check pragma value is Model
+        const fkTargetIsModel = (obj.dot(fld).getCustomPragmaVal(fk)) is Model
```
moigagoo, line 110: I think those parens are redundant.

README.rst
```diff
@@ -442,6 +441,177 @@ Norm's ``getDb`` proc lets you create a DB connection using ``DB_HOST``, ``DB_US
     ....   db.select(customerBar, "User.email = ?", "bar@bar.bar")
 
 
+Handle FOREIGN KEY manually
```
moigagoo, line 444: There are a bunch of typos but I'll fix them after merge.

src/norm/pragmas.nim
```diff
@@ -19,3 +19,80 @@ template ro* {.pragma.}
 
 template unique* {.pragma.}
   ## Mark field as unique.
+
+template fk*(val: typed) {.pragma.}
+  ##[ Mark ``int`` field as foreign key. Foreign keys always references the field ``id`` of ``val``. ``val`` should be a Model.
```
moigagoo, line 24: That's  a single line comment.

src/norm/pragmas.nim
```diff
@@ -19,3 +19,80 @@ template ro* {.pragma.}
...
+  ]##
+runnableExamples:
```
moigagoo, line 26: I used to use runnableExamples in the docstring but they turned to be too fragile. If this exact code is already in the tests, we can safely remove it from here. But I'm open to a discussion on, I can see how they are useful.

src/norm/pragmas.nim
```diff
@@ -19,3 +19,80 @@ template ro* {.pragma.}
...
+      name: string
+      productId {.fk: Product.}: int
+
+  proc newProduct(): Product=
+    Product(name: "", price: 0.0)
+
+  proc newConsumer(name: string = "", productId: int = 0): Consumer=
+    Consumer(name: name, productId: productId)
+
+  let dbName = getTempDir() / "example.db"
```
moigagoo, line 47: That's a great idea to use temp dirs! I should rewrite the tests to use it too! Thanks!

src/norm/pragmas.nim
```diff
@@ -19,3 +19,80 @@ template ro* {.pragma.}
...
+  # Clean db
+  discard tryRemoveFile(dbName)
+  let db = open(dbName, "", "", "")
+  # Depending on how sqlite3 was compiled, enabling foreign_keys may be necessary
+  db.exec(sql"PRAGMA foreign_keys = ON")
+
+  block:
```
moigagoo, line 54: I see you use those block based tests quite often, which is not in line with the rest of the tests. I'm not against that style per se but I'd rather have all tests sharing one consistent structure.

tests/tpostgresfkmodel.nim
```diff
@@ -0,0 +1,91 @@
...
+
+  teardown:
+    close dbConn
+    resetDb()
+
+  test "Create, Insert, Select with FK and Nested Models":
+    block: # Create table
+      dbConn.createTables(newBar())
+
+    block: # Insert
```
moigagoo, line 71: Again on blocks. This could be rewritten to separate test cases, this way unittest would print more detailed reports.

tests/tpostgresfkmodel.nim
```diff
@@ -0,0 +1,91 @@
...
+      var foo : Foo = Foo(a: 11, b: 12.36)
```
moigagoo, line 72: Please follow the code formatting used in other files: no space before `:`, `var` blocks when possible.

moigagoo (changes requested):
Finally, I've found the time to review your PR! It looks really well, thanks a lot. There are a few issues with code formatting and structure and an issue with `{.error.}` pragma.

Whenever I'm marking something for one backend, the same applies to another. I'm just not duplicating comments.

Please polish the PR a bit. I'll merge it, massage the code a bit myself afterwards, and cut a new release. Hopefully, this will happen within the next week.

src/norm/postgres.nim
moigagoo, line 106: Shouldn't that be `SomeInteger`? `uint` looks like a valid type for `id` but it's not `int`. It is `SomeInteger` though.

moigagoo (changes requested):

src/norm/postgres.nim
Clonkk: I did wondered if ``{.fatal: ... .}`` would have been better; in practice, I have observed the behaviour to be the same. Nonetheless, If we want to be 100% sure, I believe ``fatal`` pragma would be more appropriate.

src/norm/postgres.nim
Clonkk: I believe from the manual that it should be ``notin``:
>   notin is the same as notIn and not_in (all-lowercase version (notin, isnot) is the preferred way of writing keywords).

src/norm/pragmas.nim
Clonkk: IMO, the benefit of runnableExample is that they are compiled & executed when you use ``nim doc``.

This allows for: 
* Knowing if your examples are not correct (catches typo, things like that).
* Old obsolete example that do not work anymore do not get forgotten. Make maintenance easier.
* You can use ``nim doc`` as part of your validation process (for instance, by having a tests that execute nim doc) to make cover more case (without having to write both a test and an example).

That said, if you don't use nim doc I can just remove it.

src/norm/postgres.nim
Clonkk: I hesitated between the two. But I didn't know if you wanted to allow integer type other than ``int`` to be valid key in norm :).

tests/tsqlitefkmodel.nim
```diff
@@ -0,0 +1,80 @@
...
+    dbConn.insert(inputbar)
+    doAssert inputbar.id == 1
+    check inputbar.id == 1
+
+    var bar = newBar()
+    dbConn.select(bar, """"Bar".id = $1""", 1)
+    doAssert bar.id == 1
+    check bar.id == 1
+
+    doAssert bar.foo.a == 11
```
moigagoo, line 74: I'm not sure what your plan is with using both doAsserts and checks.

moigagoo (approved):

moigagoo:
Merging, will fix minor issues myself afterwards.

tests/tsqlitefkmodel.nim
Clonkk: That's actually something i started to fix; but then I moved to another branch to work on tests & testament

PR #96, opened 2020-11-27
fix api index and test result links

HandOfGod94:
The current links are not working.  
IMO, these are important links for any newbie like me to get their hands dirty and hence the fix


moigagoo:
@HandOfGod94 Hi! Thanks for the PR!

The links work fine when you view them in the website: https://norm.nim.town/.

Since they are relative, they don't work when you view them in the README rendered by GitHub.

On one hand, I prefer relative link as they don't break if I change the domain.

On the other hand, it's only natural to click on a link in README, they should work.

So, I'll merge the PR, thanks for the contribution.

PR #97, opened 2020-11-28
RFC: Add ability to create test transaction

HandOfGod94:
## Why this?
In other stacks (e.g. rails, elixir) provides a sandbox transaction for test env. This makes testing database interaction easily and confidently.  
The rollback after each test ensures that database remains side-effect free.

## Apporach
Create a `dbConn` which always rollbacks and inject as dependency.

Assuming `MyTask` is a valid model.
e.g.

```nim
proc listAllTasks*(dbConn: DbConn): seq[MyTask] =
  var tasks = @[newMyTask()]
  dbConn.selectAll(tasks)
  result = tasks
```

can be tested like with `dbConn.testTransaction` ?

```nim
  test "get list of all tasks":
    dbConn.testTransaction:
      var task = newMyTask("doThis", "gahan")
      dbConn.insert(task)
      let uncommittedRows = listAllTasks(dbConn)
      check(uncommittedRows.len == 1)

    check(listAllTasks(dbConn).len == 0)
```

This frees the db clean up tasks for unit testing.
> P.S. This only works if `DbConn` is injected.

### DIFF
--- a/src/norm/postgres.nim
+++ b/src/norm/postgres.nim
@@ -291,3 +291,26 @@ template transaction*(dbConn; body: untyped): untyped =
       debug rollbackQry
     dbConn.exec(sql rollbackQry)
     raise
+
+template testTransaction*(dbConn; body: untyped): untyped =
+  ##[ Wrap code in DB test transaction.
+
+  This is useful for unit testing database operations in isloation.
+
+  ]##
+
+  let
+    beginQry = "BEGIN"
+    rollbackQry = "ROLLBACK"
+
+  try:
+    when defined(normDebug):
+      debug beginQry
+    dbConn.exec(sql beginQry)
+
+    body
+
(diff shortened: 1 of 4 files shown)


moigagoo:
@HandOfGod94 sorry for the slow response.

I'm not sure I'm seeing value in this addition.

1. How is it different from transaction? The code just copies the code for transaction.
2. Are there any uses outside of writing tests? With tests, it's more reliable to ensure DB state with setup and teardown blocks.

HandOfGod94:
Yeah you are probably right. From value perspective it's not adding something really good. I was thinking of if it's possible to avoid/skip setting and cleaning up database in test env. This has kinda became norm with productive frameworks.

Closing it for now. I might need to come up with something else. This is not appropriate solution.

PR #126, opened 2022-01-04
Forbid mutating procs for types with `ro` pragma.

moigagoo:


PhilippMDoerner (commented):
This looks almost perfect!
Could the pragma have a human readable name, e.g. "readOnly", similar to how you have the `tableName` pragma? 
I know that ro stands for read-only right now since I know the context, but I had never seen that acronym in use before so I assume I'd likely find myself confused by it in 6 months (or somebody that doesn't know the context wouldn't be certain at a glance what the ro stands for)

moigagoo:
The only reason I use ro here is because it already exists :-)

However, I'm ok to deprecate ro in favor of readOnly.

moigagoo:
Imo it's better to do the pragma renaming in a separate PR. This one is about this one feature.

PR #129, opened 2022-01-06
Added starting point for caveat documentation

PhilippMDoerner:
The documentation covers:
1) How to handle simple Many-To-One relationship queries, where you can keep it all in one query by just querying from the "Many" side instead of the "One" side
2) How to handle more complex Many-To-One relationships and Many-To-Many relationships in general by making separate queries and using "Container"-objects to keep the data together
3) The mention that there is no direct support for Many-To-Many relationships and that the developers needs to create the necessary relationship-table themselves

Feel free to criticize what I've wrote here when there are places I expressed something in a confusing manner or could've done so more succinctly. I view what I wrote as merely a base starting point that we can refine as needed to make the best we can out of.

I have not added this file to wherever I'd need to add it for it to show up in the nimibook... mostly because I have no idea how to start one up locally, nor in which file I'd need to go to add it.

### DIFF
--- a/book/caveats.nim
+++ b/book/caveats.nim
@@ -6,15 +6,15 @@ nbUseNimibook
 
 nbText: """
 # Caveats
-There are some caveats when working with norm that you need to consider and strategies to work around them.
+There are some caveats when working with Norm that you need to consider and strategies to work around them.
 
-## Managing data for Many-To-Many relationships
-Support for Many-To-Many relationships has not yet been reached. You will have to set-up and manage the necessary "glue"-models yourself as if they were normal models.
+## Managing Data for Many-to-Many Relationships
+Support for Many-To-Many relationships has not yet been reached. You will have to set-up and manage the necessary "glue"-models yourself as if they were regular models.
 
-## Fetching data from Many-To-One, Many-To-Many and unidirectional One-To-One relationships
+## Fetching Data From Many-to-One, Many-to-Many and Unidirectional One-to-One Relationships
 One of the biggest caveats will be around fetching data from relationships where your model does not have a foreign key field to the other model whose data you want to query. There is no direct solution for this, but various ways to work around this.
 
-### Fetching data for simple Many-To-One/Many-To-Many relationships
+### Fetching Data for Simple Many-to-One/Many-to-Many Relationships
 Say we have a `Producer` that produces various `Products`, a classic One (Producer) to Many (Products) relationship.
 If you wanted to query the producer and all of their products in one go, you can still do that, but the other way around. Instead of querying for a producer and fetching all their products, you can query for all products of a given producer and fetch the data of the producer as shown in the Tutorial section.
 
--- a/book/caveats.nim
+++ b/book/caveats.nim
@@ -1,0 +1,81 @@
+import nimib, nimibook
+
+
+nbInit
+nbUseNimibook
+
+nbText: """
+# Caveats
+There are some caveats when working with norm that you need to consider and strategies to work around them.
+
+## Managing data for Many-To-Many relationships
+Support for Many-To-Many relationships has not yet been reached. You will have to set-up and manage the necessary "glue"-models yourself as if they were normal models.
+
+## Fetching data from Many-To-One, Many-To-Many and unidirectional One-To-One relationships
+One of the biggest caveats will be around fetching data from relationships where your model does not have a foreign key field to the other model whose data you want to query. There is no direct solution for this, but various ways to work around this.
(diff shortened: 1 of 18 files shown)


book/caveats.nim
```diff
@@ -0,0 +1,81 @@
+import nimib, nimibook
+
+
+nbInit
+nbUseNimibook
+
+nbText: """
+# Caveats
+There are some caveats when working with norm that you need to consider and strategies to work around them.
```
moigagoo, line 9: Please use consistent naming: Norm instead of norm.

It just looks tidier this way.

book/caveats.nim
```diff
@@ -0,0 +1,81 @@
...
+
+## Managing data for Many-To-Many relationships
```
moigagoo, line 11: I'm using title case in the docs, i.e. all words except for articles and prepositions start with caps: Managing Data for Many-to-Many Relationships.

book/caveats.nim
```diff
@@ -0,0 +1,81 @@
...
+Support for Many-To-Many relationships has not yet been reached. You will have to set-up and manage the necessary "glue"-models yourself as if they were normal models.
```
moigagoo, line 12: I'm not sure I should get picky with the words really, but “regular” models sounds better to me than “normal.”

moigagoo (changes requested):
I'm mostly happy with the PR. There are a few minor nitpicks but that's about it.

Also, there's “Caveats” section on the tables tutorial: https://norm.nim.town/tutorial/tables.html

Maybe it should be moved here.

book/caveats.nim
PhilippMDoerner: > I'm mostly happy with the PR. There are a few minor nitpicks but that's about it.
> 
> Also, there's “Caveats” section on the tables tutorial: https://norm.nim.town/tutorial/tables.html
> 
> Maybe it should be moved here.

I noticed that one as well. I'm perfectly happy to move it to the tables-section, though in my own opinion I'm absolutely not sure where I would put it.

The contents of it are caveats for the rows-section more than anything, as they deal for the most part with caveats around  fetching data, so DML. Given that, they seem slightly misplaced in the tables-section, which deals more in DDL.

The rows section meanwhile is already long enough imo, which makes it a tempting proposition to have a general caveats section in itself. Thus, I'm really not sure what would be the best option.

Shall we go with appending it to the tables-section anyway?

book/caveats.nim
PhilippMDoerner: Done

book/caveats.nim
PhilippMDoerner: Done

moigagoo:
Oh, my proposal was to move the caveats out of tables section to your new page, not the other way around :-)

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

PhilippMDoerner:
> Oh, my proposal was to move the caveats out of tables section to your new page, not the other way around :-)

Ohhhh now I see, yeah that's actually a really good question. One that I'm also really struggling, because I really liked how that one was handled regarding models. That being, that fundamental caveat information follows immediately after introducing the initial concept. The information given in the table section is fundamental to dealing with models. There's no way to work with Norm while not following the suggestions there.

My gut feeling would be for now to have this as a page in the tutorial, directly following the "rows" page. That way, just like with tables, you have the caveats belonging to rows following it immediately thereafter, but you also don't "overload" the rows-page. Maybe it could be called specifically `DML-Caveats` or `Row Caveats`.

But that's honestly just my gut feeling.

moigagoo:
Ok, let's leave it as it is now.

Please add the new page to nbook Nim and let's merge that PR.

PhilippMDoerner:
Alright, I fixed up some things that were stopping compilation of it and added it to nbook.nim

I still can't fully compile this thing, as it appears I receive this error after running `./nbook init` and then `./nbook build`:
`Error: unhandled exception: Error finding docs/tutorial/rowCaveats.html : no such file or directory [IOError]`
I am unsure how to interpret that one and googling didn't get me anywhere quick either. Shouldn't nimibooks build that HTML for me? What am I doing wrong here? Is this a show-stopper?

moigagoo:
@PhilippMDoerner The error occurs because there's an `nbSave` missing at the end.

PhilippMDoerner:
> @PhilippMDoerner The error occurs because there's an `nbSave` missing at the end.

Fixed and done, it compiles now on my machine

moigagoo (approved):

PR #138, opened 2022-02-21
fix #137; add tests; add changelog

### LINKED ISSUE #137: Self referencing foreign key cannot be used with ``fk`` pragma
```nim
type
  Foo[T] {.tableName: "Fooz".} = ref object of Model
    name*: string
    desc*: string
    # fooid*: Option[Foo[T]] # <- This works
    fooid* {.fk: Foo.}: Option[int64] # <- This doesn't works
```

This is due to : 
* the condition evaluation around ``hasCustomPragma(fk)`` which expects an integer (so Option[int] is forbidden)
* A type referencing itself through the fk pragma is incoplete and is not seen as a Model during type evaluation.

Clonkk:
PR Following #137

### DIFF
--- a/src/norm/sqlite.nim
+++ b/src/norm/sqlite.nim
@@ -76,7 +76,7 @@ proc createTables*[T: Model](dbConn; obj: T) =
 
   for fld, val in obj[].fieldPairs:
     if val.model.isSome:
-        dbConn.createTables(get val.model)
+      dbConn.createTables(get val.model)
 
   var colGroups, fkGroups: seq[string]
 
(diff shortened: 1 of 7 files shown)


src/norm/sqlite.nim
```diff
@@ -124,7 +130,7 @@ proc createTables*[T: Model](dbConn; obj: T) =
 
 # Row manipulation
 
-proc insert*[T: Model](dbConn; obj: var T, force = false, conflictPolicy = cpRaise) =
+proc insert*[T: Model](dbConn; obj: var T; force = false; conflictPolicy = cpRaise) =
```
moigagoo, line 133: @PhilippMDoerner are you using an autoformatter? I'm not really against semicolons but I'd rather have one style across the codebase than two ones :-)

Could you please replace semicolons with commas?

moigagoo (approved):

src/norm/sqlite.nim
Clonkk: Ah yes, I have nimpretty set up to apply automatically. I don't know why it does that, I'll fix it

PhilippMDoerner: @moigagoo not my PR :laughing:

moigagoo: @PhilippMDoerner Oops, sorry! 😬

Clonkk:
@moigagoo  Do you think it would be doable to put  a tag once this PR is done in order to specify a minimal version so we can use this feature with Nimble ?

moigagoo (approved):

moigagoo:
@Clonkk here you go, released 2.3.6.

Thanks for your contributions!

moigagoo:
@Clonkk the tests are failing. Investigating.

Clonkk:
Oh that's weird it was ok when I ran them locally

moigagoo:
> Out of curiosity why not execute the test on PR to branch devel ?

BTH I was 100% sure they do run on develop after PR. I think I just didn't let them finish, wanted to move the release faster.

Clonkk:
How do you feel about running them on the PR itself ?

PR #139, opened 2022-02-22
fix postgres tests importing sqlite; add pull request to CI

Clonkk:
* Fix mistake in https://github.com/moigagoo/norm/pull/138

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,6 +2,10 @@ name: Run Tests
 
 on:
   push
+  pull_request:
+    branches:
+      - master
+      - develop
 
 jobs:
   Tests:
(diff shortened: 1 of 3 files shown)


moigagoo:
I'm fixing it in develop directly, closing the PR.

Also, note that SQLite uses ? as the placeholder character, not $1 like Postgres.

Clonkk:
> I'm fixing it in develop directly, closing the PR.

As you wish.

> Also, note that SQLite uses ? as the placeholder character, not $1 like Postgres.

Okay, I just assumed they were interchangeable

moigagoo:
> Okay, I just assumed they were interchangeable

I think you are right. Maybe I can reuse the code that generates queries :-)

PR #141, opened 2022-03-19
Added some docs for querying nested relationships

PhilippMDoerner:
Stumbled over this one and thought documentation might be nice.
Basically I had 3 tables: Campaigns (a dnd campaign), Session (a single dnd session that is part of a campaign), Diaryentries (notes about a single session from one specific person). Session's FK-column to Campaign was called "campaign_id", diaryentries FK-column to Session was called "session_id". 

I wanted to query all Diaryentries for a campaign with a specific name, which is a nested relationship since that isn't a field directly on Session nor Diaryentry, so I had to figure out how to write that "WHERE"-condition and noticed that it's basically just concatenating the FK_fields with "_".

So I could just write the condition`session_id_campaign_id.name LIKE ?`. Given that it's not incredibly unlikely to want to do this type of query, I though adding docs for it and that norm supports it could be useful.

### DIFF
--- a/book/tutorial/rows.nim
+++ b/book/tutorial/rows.nim
@@ -97,6 +97,34 @@ nbCode:
     echo customer.user[]
 
 nbText: """
+
+If you query relationships that are nested, such as when customers can have pets and you want to query all pets of all customers of users with a specific email address, you will need  to concatenate the foreign-key fields, separeted by a `_` in your query.
+"""
+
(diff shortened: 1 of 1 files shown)


moigagoo:
Documentation improvements are always welcome :-)

Thank you so much for your PRs. They mean world to me, honestly 🙏

PR #144, opened 2022-04-20
Feature/143 better error message

PhilippMDoerner:
PR for #143 

Fixes the fact that, if you were to use `selectManyToMany` with a model like this:

```nim
type CharacterEncounterConnection* {.tableName: ENCOUNTER_CHARACTER_TABLE.} = ref object of Model
    character_id* {.fk: Character.}: int64
   encounter_id* {.fk: Encounter.}: int64

    let joinModelSeq =  @[CharacterEncounterConnection(character_id: -1, enocunter_id: -1)]
    let targetSeq = @[Encounter(title: "")]
    connection.selectManyToMany(entry, joinModelSeq, targetSeq)
```

You would get completely useless error messages such as this:

```nim
Error: type mismatch: got 'seq[int64]' for '
type
  OutType`gensym103 = typeof(
    block:
      var it: typeof(items(joinModelEntries), typeOfIter)
      it.encounter_id, typeOfProc)
block:
  let :tmp_5066724439 = joinModelEntries
  template s2_5066724440(): untyped =
    :tmp_5066724439
  
  var i`gensym103 = 0
  var result`gensym103 = newSeq(len(:tmp_5066724439))
  for it in items(:tmp_5066724439):
    result`gensym103[i`gensym103] = it.encounter_id
    i`gensym103 += 1
  result`gensym103' but expected 'seq[EncounterRead]'
```

Now you get something readable like this:

```nim
Error: unhandled exception: /home/philipp/dev/norm/src/norm/model.nim(204, 13) `joinModel()[].specialty is target` Tried using an invalid join model. Field 'BadSpecialtyJoinModel.specialty' was not of the required type `Specialty`, but of type `int64`! [AssertionDefect]
```

PR #152, opened 2022-06-21
replaced useless variable assignments

PhilippMDoerner:
Elegantbeef taught me that there is static discard.
Which is exactly what I wanted, as the validateFkField proc is solely there for its side effects that causes compile time errors if a relationship between 3 types is invalid.

This PR has the benefit of getting rid of various compiler warnings about unused variables (the tmp ones that we get rid of here).

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,11 @@
 -   [t]—test suite improvement
 -   [d]—docs improvement
 
+## 2.4.1 (Jun 23, 2022)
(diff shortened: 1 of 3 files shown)


PhilippMDoerner:
@moigagoo  I've added the changes of this PR and the other PR (#144 ) to the changelog here, in order to avoid merge conflicts when merging both branches into develop.

PR #155, opened 2022-07-08
enable deepcopy for tests

xflywind:
Hello, the Nim v2 is going to default to orc. For --gc:arc|orc 'deepcopy' support has to be enabled with --deepcopy:on. This option appears since 1.4.0

ref https://github.com/nim-lang/Nim/pull/19972

### DIFF
--- a/tests/config.nims
+++ b/tests/config.nims
@@ -1,3 +1,3 @@
 switch("path", "$projectDir/../src")
 switch("define", "normDebug")
-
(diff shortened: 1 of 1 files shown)


moigagoo:
@xflywind could you please add a changelog entry? This is a test suite improvement and needs to be recorded.

xflywind:
Sure, thank you!

PR #156, opened 2022-07-08
removed pointless variable assignment the second

PhilippMDoerner:
Miniature PR, I saw I missed this spot for using "discard" instead of a temporary variable assignment


PhilippMDoerner:
I'll do this in one larger PR, I noticed another issue

PR #157, opened 2022-07-08
Removed pointless variable assignments the third

PhilippMDoerner:
Similar as before, I missed these variable assignments which are pointless and can be easily replaced by just using "discard" or "static: discard"


PhilippMDoerner:
I'll do this in one larger PR, I noticed another issue

PR #158, opened 2022-07-09
Feature/142 multi many to many query procs

PhilippMDoerner:
There were 2 issues in my new feature:
1) There were still some variable assignments instead of using static discard
2) If you were using `selectOneToMany` and the "many"-model had the FK field as optional or just annotated via fk pragma, the proc couldn't deal with it. Similarly, `selectManyToMany` could not deal with the joinModel having an Option[Many-Model] type. (It can't deal with int64 or Option[int64] types by design since the goal is to extract the actual model from the joinModel field).

This PR should fix that. Changelog adjustments are already included. Tests all pass

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,10 @@
 -   [t]—test suite improvement
 -   [d]—docs improvement
 
+## 2.5.1 (July 9, 2022)
+- [f] Fixed further points where compile-time assertions created unused variables
+- [f] Fixed `selectOneToMany` and `selectManyToMany` that were introduced in 2.5.1 being unable to deal with FK fields that were Optionals or directly ids.
+
 ## 2.5.0 (July 7, 2022)
 
 - [+] Added `selectOneToMany` proc overload that is able to query multiple many-to-one relationships at once (see [#142](https://github.com/moigagoo/norm/issues/142))
--- a/src/norm/postgres.nim
+++ b/src/norm/postgres.nim
@@ -404,11 +404,13 @@ proc selectOneToMany*[O: Model, M: Model](dbConn; oneEntries: seq[O], relatedEnt
   for entryId in entryIds:
     let id = entryId
     when M.dot(foreignKeyFieldName) is int64:
-        relatedEntries[entryId] = relatedEntriesSeq.filter(target => target.dot(foreignKeyFieldName) == id)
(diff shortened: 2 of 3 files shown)


changelog.md
```diff
@@ -8,6 +8,11 @@
 -   [d]—docs improvement
 
 
+## 2.5.1 (July 9, 2022)
+- [f] Fixed further points where compile-time assertions created unused variables
```
xflywind, line 12:
```suggestion
- [f] Fixed further points where compile-time assertions created unused variables.
```

src/norm/postgres.nim
```diff
@@ -403,7 +403,14 @@ proc selectOneToMany*[O: Model, M: Model](dbConn; oneEntries: seq[O], relatedEnt
...
     let id = entryId
-    relatedEntries[entryId] = relatedEntriesSeq.filter(target => target.dot(foreignKeyFieldName).id == id)
+    when M.dot(foreignKeyFieldName) is int64:
+      relatedEntries[entryId] = relatedEntriesSeq.filter(target => target.dot(foreignKeyFieldName) == id)
+    elif M.dot(foreignKeyFieldName) is Option[int64]:
+      relatedEntries[entryId] = relatedEntriesSeq.filter(target => target.dot(foreignKeyFieldName).get() == id)
+    elif M.dot(foreignKeyFieldName) is Option[O]:
+      relatedEntries[entryId] = relatedEntriesSeq.filter(target => target.dot(foreignKeyFieldName).get().id == id)
+    else:
+      relatedEntries[entryId] = relatedEntriesSeq.filter(target => target.dot(foreignKeyFieldName).id == id)
```
xflywind, line 413: Imo, postgres and sqlite can the same code in a seperate file. You can probably extract it into a templates.

PhilippMDoerner: They have highly similar code, I'm struggling to split that code out properly. To just move that stuff into a template, the template would need to know about:

- M
- O
- foreignKeyFieldName
- relatedEntries
- rekatedEntriesSeq
- id/entryId

A template that takes in 6 parameters seems iffy to me.

moigagoo: Can't we use filterIt instead of filter and avoid arrow proc syntax?

PhilippMDoerner: @moigagoo : Done

PR #159, opened 2022-07-14
Escape column name

cmd410:
This is a fix for fields that match SQL keywords to cause syntax error in the query.

For example:
```nim
import std / options

import norm / [model, sqlite]

type
  User* = ref object of Model
    email*: string
    group*: int

func newUser*(email = ""): User =
  User(email: email, group: 0)

let dbConn* = open("test.db", "", "", "")
dbConn.createTables(newUser())
```
Will cause `Error: unhandled exception: near "group": syntax error [DbError]` as `group` matches an SQL keyword.
However it is possible to create such fields delimiting them with quotes:
```sql
CREATE TABLE "User"(
  "email" TEXT NOT NULL,
  "group" INTEGER NOT NULL,
  "id" INTEGER NOT NULL PRIMARY KEY
)
```
So this fix uses `strutils.escape` proc to achieve that, though not sure if its a proper way to do this...

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,9 @@
 -   [t]—test suite improvement
 -   [d]—docs improvement
 
+## 2.5.1 (July 19, 2022)
+
+- [f] Usage of SQL keywords as field names is now possible by quoting them in the resulting query.
 
 ## 2.5.0 (July 8, 2022)
 
--- a/norm.nimble
+++ b/norm.nimble
@@ -1,6 +1,6 @@
 # Package
 
-version       = "2.5.0"
+version       = "2.5.1"
 author        = "Constantine Molchanov"
 description   = "Nim ORM for SQLite and PostgreSQL."
 license       = "MIT"
--- a/norm.nimble
+++ b/norm.nimble
@@ -1,6 +1,6 @@
 # Package
 
-version       = "2.5.1"
+version       = "2.5.0"
 author        = "Constantine Molchanov"
 description   = "Nim ORM for SQLite and PostgreSQL."
 license       = "MIT"
--- a/src/norm/model.nim
+++ b/src/norm/model.nim
@@ -44,7 +44,7 @@ func table*(T: typedesc[Model]): string =
 func col*(T: typedesc[Model], fld: string): string =
   ## Get column name for a `Model`_ field, which is just the field name.
 
-  fld
+  fld.escape
 
 func col*[T: Model](obj: T, fld: string): string =
   ## Get column name for a `Model`_ instance field.
--- a/src/norm/model.nim
+++ b/src/norm/model.nim
@@ -44,7 +44,7 @@ func table*(T: typedesc[Model]): string =
 func col*(T: typedesc[Model], fld: string): string =
   ## Get column name for a `Model`_ field, which is just the field name.
 
-  fld.escape
+  "\"" & fld & "\""
 
 func col*[T: Model](obj: T, fld: string): string =
   ## Get column name for a `Model`_ instance field.
--- a/tests/common/tmodel.nim
+++ b/tests/common/tmodel.nim
@@ -16,26 +16,26 @@ suite "Getting table and columns from Model":
       pet = newPet("cat", toy)
       person = newPerson("Alice", pet)
 
-    check person.col("name") == "name"
-    check pet.col("species") == "species"
+    check person.col("name") == "\"name\""
+    check pet.col("species") == "\"species\""
 
-    check person.cols == @["name", "pet"]
-    check person.cols(force = true) == @["name", "pet", "id"]
+    check person.cols == @["\"name\"", "\"pet\""]
+    check person.cols(force = true) == @["\"name\"", "\"pet\"", "\"id\""]
 
-    check person.fCol("name") == """"Person".name"""
-    check pet.fCol("species") == """"Pet".species"""
+    check person.fCol("name") == """"Person"."name""""
+    check pet.fCol("species") == """"Pet"."species""""
 
     check person.rfCols == @[
-      """"Person".name""",
-      """"Person".pet""",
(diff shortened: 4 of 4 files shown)


cmd410:
@moigagoo yep, makes sense, fixed this

moigagoo:
Pls update the tests.

moigagoo:
@cmd410 some tests are failing.

cmd410:
truthfully, I got no clue why they fail... error messages are confusing, saying some field does not exist. The other is like `Unhandled exception: ERROR:  database "postgres" is being accessed by other users` which i don't see related to field quoting. They are all related to postgres, and I don't have one installed(it just won't install on my system for some reason) so can't even really test it, unfortunately

moigagoo:
@cmd410 most errors have to do with incorrect column names:
```
LINE 1: SELECT species, favToy, id FROM "Pet"
                        ^
HINT:  Perhaps you meant to reference the column "Pet.favToy".
```
```
Unhandled exception: ERROR:  column "userid" referenced in foreign key constraint does not exist
```

These should be easy to fix. It's either an error in the tests or in query generation for e.g. foreign key queries.

The `database "postgres" is being accessed by other users` error is probably caused by incorrectly closed DB connections. You can ignore them, they're likely to go away when you fix the column names.

PhilippMDoerner:
In order to help speed this along I was also taking a look at the tests and I'm a bit confused by what's supposed to be wrong.
Running e.g. `tdbtypes` with  `sudo docker-compose run --rm test tests/postgres/tdbtypes.nim` gives you this:
```txt
+[Suite] Import dbTypes from norm/private/postgres/dbtypes
+/usr/src/app/tests/postgres/tdbtypes.nim(36) tdbtypes
+/usr/src/app/src/norm/postgres.nim(214) select
+/root/.nimble/pkgs/ndb-0.19.9/ndb/postgres.nim(261) getAllRows
+/root/.nimble/pkgs/ndb-0.19.9/ndb/postgres.nim(220) dbError
+
+    Unhandled exception: ERROR:  column "lastlogin" does not exist
+LINE 1: ...User"."lastLogin", "User"."id" FROM "User"  WHERE lastLogin ...
+                                                             ^
+HINT:  Perhaps you meant to reference the column "User.lastLogin".
+ [DbError]
+  [FAILED] dbValue[DateTime] is imported
```

Since I couldn't get the test to output its queries I copy pasted the code into a playground project and ran it directly like this:
```nim
import std/[unittest, os, strutils, times, sugar, with, logging]
import norm/[model, sqlite, types]

addHandler(newConsoleLogger(levelThreshold = lvlDebug))

type
  User* = ref object of Model
    lastLogin*: DateTime

  Number* = ref object of Model
    n*: int
    n16*: int16
    n32*: int32
    n64*: int64

  String* = ref object of Model
    s*: string
    sc10*: StringOfCap[10]
    psc5*: PaddedStringOfCap[5]

proc newUser*(): User = User(lastLogin: now())

func `===`*(a, b: User): bool =
  a.lastLogin == b.lastLogin

func newNumber*(n: int, n16: int16, n32: int32, n64: int64): Number =
  Number(n: n, n16: n16, n32: n32, n64: n64)

func newNumber*: Number =
  newNumber(0, 0'i16, 0'i32, 0'i64)

func `===`*(a, b: Number): bool =
  a[] == b[]

func newString*(s: string, sc10: StringOfCap[10], psc5: PaddedStringOfCap[5]): String =
  String(s: s, sc10: sc10, psc5: psc5)

func newString*: String =
  newString("", StringOfCap[10]"", PaddedStringOfCap[5]"")

func `===`*(a, b: String): bool =
  a[] == b[]

let dbConn = open(":memory:", "", "", "")
dbConn.createTables(newUser())
dbConn.createTables(newNumber())
dbConn.createTables(newString())

let users = @[newUser()].dup:
  dbConn.select("""lastLogin <= ?""", ?now())

echo users.repr
check len(users) == 0
```

When I run the code above locally, I receive not a single error.
The queries it generates are perfectly also looking perfectly valid:
```txt
DEBUG CREATE TABLE IF NOT EXISTS "User"("lastLogin" FLOAT NOT NULL, "id" INTEGER NOT NULL PRIMARY KEY)
DEBUG CREATE TABLE IF NOT EXISTS "Number"("n" INTEGER NOT NULL, "n16" INTEGER NOT NULL, "n32" INTEGER NOT NULL, "n64" INTEGER NOT NULL, "id" INTEGER NOT NULL PRIMARY KEY)
DEBUG CREATE TABLE IF NOT EXISTS "String"("s" TEXT NOT NULL, "sc10" TEXT NOT NULL, "psc5" TEXT NOT NULL, "id" INTEGER NOT NULL PRIMARY KEY)
DEBUG SELECT "User"."lastLogin", "User"."id" FROM "User"  WHERE lastLogin <= ? <- [1660589752.379417]
```

I tested queries similarly styled to `SELECT "User"."lastLogin", "User"."id" FROM "User"  WHERE lastLogin <= ?` (All column and tablenames in quotes except for those in the WHERE statement) and found no issue there either.
So I'm not quote sure what's going on either.

PhilippMDoerner:
I've taken a look at the errors here as well and the issue this opens up seems non trivial.
For example:
`tests/postgres/tdbtypes.nim` will only function if you replace line 37
old: `dbConn.select("""lastLogin <= $1""", ?now())`
new: `dbConn.select(""""lastLogin" <= $1""", ?now())`

Notice the extra quotation marks around the fields.
For some reason now they're *required* around the SQL the user passes in as well? Is this behaviour that we want?

PhilippMDoerner:
Given that the issues of this PR were solved in #168 I'd argue this PR belongs closed.
Further, #168 comes with its completely own set of problems that this PR would also run into, so that makes it more sensible to contain any commentary in that PR rather than this one.

Issue #133, opened 2022-01-23
[FeatureRequest] Add a public "fromRow"-like proc that allows handing a raw SQL statement and only do conversion to any object type


PhilippMDoerner:
Heyho!

As part of the web application that I am writing I am utilizing sqlite's full text search feature FTS5. It's really useful and provides basic search utilities which means you can do somewhat elaborate searches without needing to add something heavy like elasticsearch.
However, that table looks in DDL something like this:

```nim
CREATE VIRTUAL TABLE search_article_content USING fts5( title, title_rev, body, body_rev, table_name, record_id, campaign_id, guid )
```

And I want to query content on that table with an SQL Query like this 
```sql
            SELECT 
                table_name,
                record_id,
                title,
                body,
                guid,
                bm25(search_article_content, 15) as search_score -- 15 states that a match in the title is 15 times as valuable as a match in the body.
            FROM search_article_content 
            WHERE
                campaign_id = %s
                AND (
                    title MATCH %s
                    OR title_rev MATCH %s
                    OR body MATCH %s
                    OR body_rev MATCH %s
                )
            ORDER BY search_score
            LIMIT %s;
```

All I want from norm at this point, is the functionality of taking my very custom SQL statement, make a db query with it and transforming the data it receives into an object whose type I give it. Not even an object that inherits from `Model` really, since I don't want to infer anything from the model, it's job should solely be to be the transformation target as everything else should be determined by my raw sql query. 

The proc I imagine could have a signature like this:

`rawSelect(con: DbConn, query: SqlQuery, targetEntries: seq[T])` 

T in this scenario explicitly only needs to be an object, nothing more.

Would such a feature even be desirable for norm?

PhilippMDoerner:
Correct me if I'm wrong @moigagoo but given how tightly norm interacts with models, I think conversion of Row to Model is only possible when norm generated the query, correct? So I don't think the conversion of `Row` to a `Model` instance can be used to allow raw Sql to be executed and converted into any object type.

Thus I'm getting the feeling that, if we want such a proc, it would need its own "convert to object" implementation instead of hooking into anything norm has already written.

For my own needs, I have currently solved this by using [`nisane`](https://github.com/enthus1ast/nisane), which tries to do exactly that (albeit some wonky behaviour and its very macro heavy). For something with nisane's API, I can even provide some procs on how I work with them:

```nim
import std/[db_sqlite]
import nisane 
import genericNisaneRepository

export nisane
export sequtils

[...]

PR #164, opened 2022-08-12
Feature/#133 add public from row

PhilippMDoerner:
This proc adds the feature requested by #133 :
A way to interact with the database with arbitrary SQL, yet still benefit from norms incredibly useful parsing of Row to objects, just this time it allows for arbitrary object types instead of using models.

This has the following benefits:
- It becomes possible to interact with views which do not have an `id` column, as the user can define their own types for interacting with the view
- It becomes possible to use norm for executing arbitrary complex SQL such as recursive queries, which was not possible before

The specific procs that allows this are:
- A modification of `fromRowPos` and `fromRow` which now can deal with *any ref object*  as opposed to only `Model` types
- `rawSelect` procs that make use of that and allow the user to pass non-Model ref object and a full SQL query to execute

**A small side-benefit:**
I observed around a 10% speed-up for my limited test-cases when executing the "new" version of `fromRowPos`. The reason for this speed-up is that some computation has been moved to compile-time instead of runtime.

**Things that would still need to be done:**
- General review + Approval that this feature is wanted
- ~~Figure out in which module procs such as `isContainer`, `toOptional` and `isEmptyColumn` should be stored, as `rowutils` does not feel like the appropriate place for these~~ (I've decided to put them into a shared `utils.nim` module in private)
- ~~Figure out actually decent names for `isContainer` and `toOptional`. These names are (in my mind) only placeholders since I couldn't think of a better name for now. They are more "open" alternatives to `isModel` and `model`. With the new context of using ref object instead of Model however `isRefObject` and `refObject` as names seemed confusing to me.~~ (I've decided to go with `toOptional` and `isRefObject`)
- ~~Write tests for the `rawSelect` procs (and maybe the modified `fromRowPos` proc)  ( @moigagoo I would proceed with this point and the ones below once we ironed out the other ones)~~ (Basic tests are written. The details of the parsing into ref object do not need to be tested - the parsing from object to model already does that) 
- ~~Add changelog changes~~ (DONE)
- Add nimibook docs

### DIFF
--- a/src/norm/private/postgres/rowutils.nim
+++ b/src/norm/private/postgres/rowutils.nim
@@ -6,6 +6,7 @@ import ndb/postgres
 
 import dbtypes
 import ../dot
+import ../utils
 import ../../model
 import ../../pragmas
 
@@ -14,16 +15,9 @@ when (NimMajor, NimMinor) <= (1, 6):
 else:
   import std/macros
 
-
-func isContainer*[T](val: typedesc[T]): bool {.compileTime.} = T is ref object
-func isContainer*[T](val: T): bool {.compileTime.} = T is ref object
-func isContainer*[T](val: typedesc[Option[T]]): bool {.compileTime.} = T is ref object  
-func isContainer*[T](val: Option[T]): bool {.compileTime.} = T is ref object  
-
-func toOptional*[T: ref object](val: T): Option[T] = some val
-func toOptional*[T: ref object](val: Option[T]): Option[T] = val
-func isEmptyColumn(row: Row, index: int): bool = row[index].kind == dvkNull
-
+func isEmptyColumn*(row: Row, index: int): bool =
+  ## Checks whether the column at a given index is empty
+  row[index].kind == dvkNull
 
 ## This does the actual heavy lifting for parsing
 proc fromRowPos[T: ref object](obj: var T, row: Row, pos: var Natural, skip: static bool = false) =
@@ -33,13 +27,13 @@ proc fromRowPos[T: ref object](obj: var T, row: Row, pos: var Natural, skip: sta
   ]##
 
   for fld, dummyVal in T()[].fieldPairs:
-    when isContainer(typeof(dummyVal)):                 ## If we're dealing with a ``Model`` field
+    when isRefObject(typeof(dummyVal)):                 ## If we're dealing with a ``Model`` field
       if dot(obj, fld).toOptional().isSome:             ## and it's either a ``some Model`` or ``Model``
         var subMod = dot(obj, fld).toOptional().get()   ## then we try to populate it with the next ``row`` values.
 
         if row.isEmptyColumn(pos):                      ## If we have a ``NULL`` at this point, we return an empty ``Model``:
           when typeof(dummyVal) is Option:              ## ``val`` is guaranteed to be either ``Model`` or an ``Option[Model]`` at this point,
-            when isContainer(dummyVal): 
+            when isRefObject(dummyVal):
               dot(obj, fld) = none typeof(subMod)       ## and the fact that we got a ``NULL`` tells us it's an ``Option[Model]``,
 
           inc pos
--- a/src/norm/private/postgres/rowutils.nim
+++ b/src/norm/private/postgres/rowutils.nim
@@ -14,40 +14,54 @@ when (NimMajor, NimMinor) <= (1, 6):
 else:
   import std/macros
 
-proc fromRowPos[T: Model](obj: var T, row: Row, pos: var Natural, skip = false) =
+
+func isContainer*[T](val: typedesc[T]): bool {.compileTime.} = T is ref object
+func isContainer*[T](val: T): bool {.compileTime.} = T is ref object
+func isContainer*[T](val: typedesc[Option[T]]): bool {.compileTime.} = T is ref object  
+func isContainer*[T](val: Option[T]): bool {.compileTime.} = T is ref object  
+
+func toOptional*[T: ref object](val: T): Option[T] = some val
+func toOptional*[T: ref object](val: Option[T]): Option[T] = val
+func isEmptyColumn(row: Row, index: int): bool = row[index].kind == dvkNull
+
+
+## This does the actual heavy lifting for parsing
+proc fromRowPos[T: ref object](obj: var T, row: Row, pos: var Natural, skip: static bool = false) =
   ##[ Convert ``ndb.sqlite.Row`` instance into `Model`_ instance, from a given position.
 
   This is a helper proc to convert to `Model`_ instances that have fields of the same type.
   ]##
 
-  for fld, val in obj[].fieldPairs:
-    if val.isModel:                                 ## If we're dealing with a ``Model`` field
-      if val.model.isSome:                          ## and it's either a ``some Model`` or ``Model``
-        var subMod = get val.model                  ## then we try to populate it with the next ``row`` values.
+  for fld, dummyVal in T()[].fieldPairs:
+    when isContainer(typeof(dummyVal)):                 ## If we're dealing with a ``Model`` field
+      if dot(obj, fld).toOptional().isSome:             ## and it's either a ``some Model`` or ``Model``
+        var subMod = dot(obj, fld).toOptional().get()   ## then we try to populate it with the next ``row`` values.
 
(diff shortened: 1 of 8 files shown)


src/norm/private/postgres/rowutils.nim
```diff
@@ -14,40 +14,54 @@ when (NimMajor, NimMinor) <= (1, 6):
...
+
+func isContainer*[T](val: typedesc[T]): bool {.compileTime.} = T is ref object
+func isContainer*[T](val: T): bool {.compileTime.} = T is ref object
+func isContainer*[T](val: typedesc[Option[T]]): bool {.compileTime.} = T is ref object  
+func isContainer*[T](val: Option[T]): bool {.compileTime.} = T is ref object  
+
+func toOptional*[T: ref object](val: T): Option[T] = some val
+func toOptional*[T: ref object](val: Option[T]): Option[T] = val
+func isEmptyColumn(row: Row, index: int): bool = row[index].kind == dvkNull
+
```
PhilippMDoerner, line 26: As stated, the procs above have 2 problems:
1) I don't think they belong in rowutils, I'm just not sure where else to put them. Ideally something that is shared between postgres and sqlite packages since they are **identical** between the packages
2) I can't think of better names for `isContainer` and `toOptional`. If anyone has any, please suggest.

PhilippMDoerner:
@moigagoo before I do the (not that inconsiderable amount of) work of getting this PR into a "finished" state, I *would* like your confirmation first that the feature is actually wanted.
I'd like to avoid a scenario where I bring this to the finish line, only for us to reject the PR.

tests/postgres/trawselect.nim
```diff
@@ -0,0 +1,108 @@
...
+    var mydoc = MyDoc()
+    dbConn.rawSelect(sql, mydoc, someDoctor.id)
+
+    check mydoc.name == "Vet1"
+
+  test "rawSelect multiple entries":
+    type MyPerson = ref object
+      name: string
+
+    let sql = """SELECT name FROM "Person" ORDER BY id"""
```
PhilippMDoerner, line 83: I've added some basic tests now and updated the changelog.

Sidenote that I noticed in that regard @moigagoo : 
When using `rawSelect` with postgres, the query would not recognize the table name without the `"`.
As in, this would not work: `SELECT name FROM Person ORDER BY id` 
But this would: `SELECT name FROM "Person" ORDER BY id`
This is only this way in **postgres**

Looking at the table names via
```sql
SELECT table_name FROM information_schema.tables WHERE table_schema='public' AND table_type='BASE TABLE';
```
Revealed nothing suspicious that could hint as to why the quotation marks are mandatory:
```txt
+@[@[Toy], @[Pet], @[Person], @[DoctorVisit], @[Doctor], @[PlayfulPet]]
```

Any ideas?

moigagoo: I think it might have to do with capitalized table names: https://stackoverflow.com/questions/43111996/why-postgresql-does-not-like-uppercase-table-names

Quotes should be used everywhere for consistency in both Postgresl and SQLite.

PhilippMDoerner: If that is how it is intended to be, then I should document this behaviour in the doc comment of the proc.
The most recent commit just did that.
Since SQLite does not display this requirement behaviour I did not add that same info to the sqlite proc.

PhilippMDoerner:
Since the actual conversion of `Row -> ref object` is implicitly covered by all the tests that convered from `Row -> Model` I personally think that no further tests are required. Just that they generally can work, which the `trawselect.nim` test files cover.

**To me** the work on this PR is thus done and it could be merged.

PhilippMDoerner:
@moigagoo 
Yo, it's been near a month since creation of the PR with request for comment and almost a week since it finished (which I've done to have a modified branch of norm that I can now use in my sideproject).

Is this to be on ice until the escape column name PR is merged or is something else going on?

src/norm/private/postgres/rowutils.nim
```diff
@@ -14,40 +15,47 @@ when (NimMajor, NimMinor) <= (1, 6):
...
   ##[ Convert ``ndb.sqlite.Row`` instance into `Model`_ instance, from a given position.
 
   This is a helper proc to convert to `Model`_ instances that have fields of the same type.
   ]##
 
-  for fld, val in obj[].fieldPairs:
-    if val.isModel:                                 ## If we're dealing with a ``Model`` field
-      if val.model.isSome:                          ## and it's either a ``some Model`` or ``Model``
-        var subMod = get val.model                  ## then we try to populate it with the next ``row`` values.
+  for fld, dummyVal in T()[].fieldPairs:
```
moigagoo, line 29: @PhilippMDoerner T() will not work for types that don't have default values for all.their fields.

moigagoo (commented):
There's just one comment I have. Otherwise, I'm ready to merge.

moigagoo:
@PhilippMDoerner sorry for the slow responses. I've looked through the OR and found just one place where I have concerns. Please review this comment and we'll merge the PR.

src/norm/private/postgres/rowutils.nim
PhilippMDoerner: Can you maybe provide an example?
I've been wrecking my brain over a possible type that could break this and came up empty.
Not even having a field with the `DateTime` type would break this, since I never actually access the value inside the variable. I only ever check their types, which appears to get around such issues.

PhilippMDoerner:
@moigagoo 
I've found a new reason for concern that is related to your comment but in a different manner.
It turns out there are 2 pragmas that we would be unable to support with this:
The above makes it essentially impossible to use the `requiresInit` pragma with a model/ref object.
To me that is an acceptable sacrifice, but I thought I should bring it up since it came up in convo with Yardanico.

src/norm/private/postgres/rowutils.nim
PhilippMDoerner: I got one example courtesy of mratsim.
Any field of type `range(1..2)` (Or any other range that does not include null) will not work with this.
That... I'm not sure that is fixable. To iterate at compile time you need to be able to initialize at compile-time, that's pretty much where the 10% speed up stems from. And avoiding a whole bunch of procs just to please the compiler for no benefit.

moigagoo: > Not even having a field with the DateTime type would break this

This is the example I was going to provide :-)

I'm afk now but I could swear T() would not compile if T is a ref object with a field of DateTime.

Also, since this proc accepts an instance anyway, why not use it? Why create a new one?

PhilippMDoerner: That is because I went in with the goal to only compute at runtime what absolutely must be computed at runtime and move everything else to compile-time.
`T()` gives me an instance that exists at compile-time.  It is impossible to do the same with the object that is being passed because you only know that value at runtime. Strictly speaking `T()` *should* not be necessary, in terms of information it should suffice to just have an iterator over a type that spits out its field names and types, but nim doesn't have that so a compile-time instance is required. 

I make use of the for loop being compile-time in my various `when` statements in the code, these check at compile time whether the field you need to be parsing columns into is a primitive or a ref object or an optional of a primitive/ref object. Depending on that only the needed code for parsing this particular field is added to the loop.

This way you're achieving 2 things:
1) Computation gets outsourced into compile-time that you no longer have to do at runtime, speeding up the conversion slightly (which, as stated for my limited testing with models of 2-4 fields was roughly 10% faster than before)
2) No longer are procs needed only to please the compiler because invalid code-paths no longer exist at runtime. So the proc `func to*(dbVal; T: typedesc[Model]): T =` That you had to put in because the code-path that you land in line ` val = row[pos].to(typeof(val)) ` with val being a Model type exists, even though if-statements prevent that from ever being the case.

PhilippMDoerner: > This is the example I was going to provide :-)

At least those I can guarantee you they work. Not only do I have plenty of fields in my side-project that work flawlessly with this version of norm (I'm currently using my own branch there since I wanted the `rawSelect` procs), the tests themselves would not work if DateTime objects caused a problem. E.g. `trelated.nim` contains testcases with models that have DateTime fields.

moigagoo:
I'll release 2.5.2 Monday. I'm afk for the entire weekend.

Issue #165, opened 2022-08-14
[Bugfix] Enable creation of tables whose fieldName otherwise clashes with SQL keywords such as `group`


PhilippMDoerner:
In the nim discord ajusa previously encountered this issue where they had a database, but one of the columns they had was called `group`.
This causes issues with sqlite's SQL (and likely postgres but I didn't test that). In the language you can avoid those by escaping the name with `'`.

So minimal reproducible example:
```nim
import norm/[model, sqlite]

type A = ref object of Model
    group: string 

let x = A(group: "")
let con = open(":memory:", "", "", "")
con.createTables(x)
```
```txt
Error: unhandled exception: near "group": syntax error [DbError]
```

The SQL generated is 
```sql
CREATE TABLE IF NOT EXISTS "A"(group TEXT NOT NULL, id INTEGER NOT NULL PRIMARY KEY)
```

To fix this, the SQL it would need to generate would be this (can also use `"` instead of `'`):
```sql
CREATE TABLE IF NOT EXISTS "A"('group' TEXT NOT NULL, id INTEGER NOT NULL PRIMARY KEY)
```

A similar issue goes for `selectAll` as it generates for example this sql:
```nim
import norm/[model, sqlite]

type A* = ref object of Model
    group*: string 

let con = open(":memory:", "", "", "")

let y = """CREATE TABLE IF NOT EXISTS "A"('group' TEXT NOT NULL, id INTEGER NOT NULL PRIMARY KEY)"""
con.exec(sql y)

var objs = @[A(group: "")]
con.selectAll(objs)
```
```sql
SELECT "A".group, "A".id FROM "A"  WHERE 1
```

Should be generating
```sql
SELECT "A"."group", "A"."id" FROM "A"  WHERE 1
```

moigagoo:
There's an ungoing PR that fixes this by adding quotes around column names.

The PR stuck on test suite update: since the table names are all different now, the entire test suite needs revision and update.

PR #168, opened 2022-08-22
Escape column name

PhilippMDoerner:
This is a copy of #159  but with fixes applied by me to get the tests running.
@cmd410 @moigagoo 

The main reason I didn't commit them onto the branch of @cmd410 is because I actually don't know how ^^'

For the general description, refer to #159 and a related userstory #165 

GENERAL REMARK:
I noticed some worrying behaviour which I think might cause this PR to be one that break backwards compatibility.
A fair amount of the tests in postgres did not run because they either:

1) Displayed case-sensitive naming which they did not before (The tests where this was the case have that remarked in their commit message)
2) Displayed the **requirement** that specific columns (I think only FK columns?) be surrounded by quotation marks

That is new behaviour and a breaking change. It might of course also just be that I ran the tests under linux and this might be fine under windows, but it *is* something to take a look at either way.

### DIFF
--- a/src/norm/postgres.nim
+++ b/src/norm/postgres.nim
@@ -101,7 +101,7 @@ proc createTables*[T: Model](dbConn; obj: T) =
       colShmParts.add "UNIQUE"
 
     if val.isModel:
-      var fkGroup = """FOREIGN KEY("$#") REFERENCES $#($#)""" %
+      var fkGroup = """FOREIGN KEY($#) REFERENCES $#($#)""" %
         [obj.col(fld), typeof(get val.model).table, typeof(get val.model).col("id")]
 
       when obj.dot(fld).hasCustomPragma(onDelete):
--- a/src/norm/postgres.nim
+++ b/src/norm/postgres.nim
@@ -101,7 +101,7 @@ proc createTables*[T: Model](dbConn; obj: T) =
       colShmParts.add "UNIQUE"
 
     if val.isModel:
-      var fkGroup = "FOREIGN KEY($#) REFERENCES $#($#)" %
+      var fkGroup = """FOREIGN KEY("$#") REFERENCES $#($#)""" %
         [obj.col(fld), typeof(get val.model).table, typeof(get val.model).col("id")]
 
       when obj.dot(fld).hasCustomPragma(onDelete):
@@ -120,9 +120,9 @@ proc createTables*[T: Model](dbConn; obj: T) =
           const selfTableName = '"' & T.getCustomPragmaVal(tableName) & '"'
         else:
           const selfTableName = '"' & $T & '"'
-        fkGroups.add "FOREIGN KEY ($#) REFERENCES $#(id)" % [fld, selfTableName]
+        fkGroups.add """FOREIGN KEY ("$#") REFERENCES $#(id)""" % [fld, selfTableName]
       else:
-        fkGroups.add "FOREIGN KEY ($#) REFERENCES $#(id)" % [fld, (obj.dot(fld).getCustomPragmaVal(fk)).table]
+        fkGroups.add """FOREIGN KEY ("$#") REFERENCES $#(id)""" % [fld, (obj.dot(fld).getCustomPragmaVal(fk)).table]
 
     colGroups.add colShmParts.join(" ")
 
--- a/tests/postgres/tdbtypes.nim
+++ b/tests/postgres/tdbtypes.nim
@@ -4,7 +4,6 @@ import norm/[model, postgres, types]
 
 import ../models
 
-
 const
   dbHost = "postgres"
   dbUser = "postgres"
@@ -33,7 +32,7 @@ suite "Import dbTypes from norm/private/postgres/dbtypes":
 
   test "dbValue[DateTime] is imported":
     let users = @[newUser()].dup:
-      dbConn.select("""lastLogin <= $1""", ?now())
+      dbConn.select(""""lastLogin" <= $1""", ?now())
 
     check len(users) == 0
 
--- a/tests/postgres/tfkpragma.nim
+++ b/tests/postgres/tfkpragma.nim
@@ -70,6 +70,6 @@ suite "``fk`` pragma":
     for inpCustomer in inpCustomers.mitems:
       dbConn.insert inpCustomer
 
-    dbConn.select(outCustomers, """"userid" = $1""", userA.id)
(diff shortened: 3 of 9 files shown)


tests/postgres/ttables.nim
```diff
@@ -56,7 +56,7 @@ suite "Table creation":
 
     check dbConn.getAllRows(qry, "FurnitureTable") == @[
       @[?"id", ?"bigint"],
-      @[?"legcount", ?dftDbInt]
+      @[?"legCount", ?dftDbInt]
```
PhilippMDoerner, line 59: An example for sudden case-sensitivity, though that is when comparing column names and not during actual queries, not sure how much that matters.

tests/postgres/ttables.nim
```diff
@@ -80,7 +80,7 @@ suite "Table creation":
     ]
 
     check dbConn.getAllRows(qry, "Pet") == @[
-      @[?"favtoy", ?"bigint"],
+      @[?"favToy", ?"bigint"],
```
PhilippMDoerner, line 83: An example for sudden case-sensitivity, though that is when comparing column names and not during actual queries, not sure how much that matters.

tests/postgres/trows.nim
```diff
@@ -219,7 +220,7 @@ suite "Row CRUD":
 
     let
       personRow = get dbConn.getRow(sql"""SELECT name, pet, id FROM "Person" WHERE id = $1""", person.id)
-      petRow = get dbConn.getRow(sql"""SELECT species, favToy, id FROM "Pet" WHERE id = $1""", pet.id)
+      petRow = get dbConn.getRow(sql"""SELECT species, "favToy", id FROM "Pet" WHERE id = $1""", pet.id)
```
PhilippMDoerner, line 223: An example for suddenly required quotation marks, necessary only on the FK field "favToy", not on the others

tests/postgres/tfkpragma.nim
```diff
@@ -70,6 +72,6 @@ suite "``fk`` pragma":
     for inpCustomer in inpCustomers.mitems:
       dbConn.insert inpCustomer
 
-    dbConn.select(outCustomers, """"userid" = $1""", userA.id)
+    dbConn.select(outCustomers, """"userId" = $1""", userA.id)
```
PhilippMDoerner, line 75: An example for sudden case-sensitivity, though that is when comparing column names and not during actual queries, not sure how much that matters.

tests/postgres/tdbtypes.nim
```diff
@@ -33,7 +32,7 @@ suite "Import dbTypes from norm/private/postgres/dbtypes":
 
   test "dbValue[DateTime] is imported":
     let users = @[newUser()].dup:
-      dbConn.select("""lastLogin <= $1""", ?now())
+      dbConn.select(""""lastLogin" <= $1""", ?now())
```
PhilippMDoerner, line 35: An example for suddenly required quotation marks, this time not on an FK field as lastLogin is of type DateTime. Not sure what that is about.

src/norm/postgres.nim
```diff
@@ -123,9 +123,9 @@ proc createTables*[T: Model](dbConn; obj: T) =
           const selfTableName = '"' & T.getCustomPragmaVal(tableName) & '"'
         else:
           const selfTableName = '"' & $T & '"'
-        fkGroups.add "FOREIGN KEY ($#) REFERENCES $#(id)" % [fld, selfTableName]
+        fkGroups.add """FOREIGN KEY ("$#") REFERENCES $#(id)""" % [fld, selfTableName]
       else:
-        fkGroups.add "FOREIGN KEY ($#) REFERENCES $#(id)" % [fld, (obj.dot(fld).getCustomPragmaVal(fk)).table]
+        fkGroups.add """FOREIGN KEY ("$#") REFERENCES $#(id)""" % [fld, (obj.dot(fld).getCustomPragmaVal(fk)).table]
```
PhilippMDoerner, line 128: I had to add quotation marks here because otherwise the FK constraints wouldn't have any.
Please note that I ONLY added the quotation marks in these two lines.
There is a third line dealing with FK constraints, line 107 further up, where I DID NOT add quotation marks around the field,  because when that line gets called by tests the field suddenly already has quotationmarks (possibly because it calls the col proc where the others don't).

tests/postgres/trows.nim
moigagoo: What if you put the rest in quotations for consistency? Does it break the test?

This is so weird 🤔 Possibly, this has to do with mixing cases. If the column name uses mixed case, it must be quoted. But that's just a wild guess.

src/norm/postgres.nim
moigagoo: > when that line gets called by tests the field suddenly already has quotationmarks (possibly because it calls the col proc where the others don't).

Oof, that's dangerous behavior. col proc should be called always.

tests/postgres/trows.nim
PhilippMDoerner: Changing the line to this:  `sql"""SELECT "species", "favToy", "id" FROM "Pet" WHERE id = $1"""`
Still works perfectly fine.

An observation I made though, is that this:
`sql"""SELECT "sPecies", "favToy", "id" FROM "Pet" WHERE id = $1"""`
Will cause the test to fail while this:
`sql"""SELECT sPecies, "favToy", "id" FROM "Pet" WHERE id = $1"""`
works perfectly fine.

I'm getting the impression that not only are suddenly the quotation marks sometimes mandatory, they cause the column names to suddenly be case-sensitive.

src/norm/postgres.nim
PhilippMDoerner: Just tried it, it does work.
I managed to replace `fld` on there with `obj.col(fld)` and it ran through all tests.
I'm not sure if the code that generates is correct, as stated I haven't read through this code in any intensity in order to understand what's going on, but it does pass the tests.

PhilippMDoerner:
I'm perfectly fine with leaving this PR in the air as it is, as I'm not sure myself what to do with this.
On the one hand it fixes an issue about you being unable to have certain table names that overlap with SQL keywords, on the other it introduces breaking changes and unusual behaviour for postgres users which need to put quotation marks around fields all of a sudden (and that even only sometimes).

Is this grounds for not accepting the PR?
Should the PR be accepted and we add warnings to the docs for postgres users specifically?

Issue #173, opened 2022-09-12
db.insert(model, force = true) doesn't work as expected


Clonkk:
Inserting a model with a specific ID doesn't work. 

I added this tests to ``trows.nim`` and it fails. I was expecting it to pass.

```nim
  test "Insert with forced id":
    var toy = newtoy(137.45)
    toy.id = 134
    print(toy)
    dbConn.insert(toy, force = true)
    check toy.id == 134
    dbConn.delete(toy)
```

~~I think this is a regression as we talked about that already in~~ https://github.com/moigagoo/norm/issues/104/

Okay, not a regression, I guess we misunderstood each other. What I meant with insertion with id at non-zero was not just forcing the insertion but forcing the insertion **with the given id**. 

I see from   test ``"Insert row twice"`` in ``trows`` that it was not the intended behaviour. If it's fine with you, I can open a PR changing ``force`` behavior when an id is present to insert with the given id field and not just performing the insertion and changing the object id. Conflict are already handled through the conflictPolicy

Clonkk:
@moigagoo  What do you think ? If you're okay with that, I can open the PR in the afternoon

Clonkk:
We can discuss changes here https://github.com/moigagoo/norm/pull/174

PR #174, opened 2022-09-12
force =true with obj.id != 0 in insert now use the given id

Clonkk:
This PR solves https://github.com/moigagoo/norm/issues/173


Clonkk:
The logic of forcing ids seems to be broken for postgres (I guess it handles the id increment differently than SQLite).

moigagoo:
> @moigagoo I used conflictPolicy & unified the behavior between SQLite & Postgres regarding manually inserted ids.

@Clonkk I think the reason I only implemented this feature for SQLite initially is that Postgres didn't have this at all.

Apparently, this works now as the tests pass, so I'm happy to merge this.

Thanks!

PR #177, opened 2022-10-15
Connection pool

moigagoo:
Alternative approach to https://github.com/moigagoo/norm/pull/175

This is an attempt to come up with a simpler, more explicit pool implementation. The general concept: https://github.com/moigagoo/norm/pull/175#issuecomment-1250126017

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -8,12 +8,18 @@
 -   [d]—docs improvement
 
 
+## 2.5.3 (WIP)
+
+- [+] App connection pool (see [#50](https://github.com/moigagoo/norm/issues/50)).
+
+
 ## 2.5.2 (September 14, 2022)
 
 - [+] Added `rawSelect` proc, which allows you execute raw SQL and have the output be parsed into a custom object-type.
 - [+] `insert` with `obj.id != 0` and `force=true` now uses the id provided. For PostgreSQL, this means adding conflictPolicy with `ON CONFLICT DO...` clause.
 - [r] Logging: refactored `log` module to not trigger warnings when `normDebug` is not defined.
 - [r] Slightly changed how objects are being parsed, leading to a small performance increase.
+
 
 ## 2.5.1 (July 20, 2022)
 - [+] Added `uniqueGroup` pragma to provide UNIQUE constraint on multiple columns (see [#136](https://github.com/moigagoo/norm/issues/136)).
--- a/changelog.md
+++ b/changelog.md
@@ -8,7 +8,7 @@
 -   [d]—docs improvement
 
 
-## 2.5.3 (WIP)
+## 2.6.0 (WIP)
 
 - [+] App connection pool (see [#50](https://github.com/moigagoo/norm/issues/50)).
 
--- a/src/norm/pool.nim
+++ b/src/norm/pool.nim
@@ -1,12 +1,13 @@
 import std/locks
 
-import sqlite
+import sqlite, postgres
 
 
 type 
-  Pool* = ref object 
+  Conn = sqlite.DbConn | postgres.DbConn
+  Pool*[T: Conn] = ref object 
     defaultSize: Natural
-    conns: seq[DbConn]
+    conns: seq[T]
     poolExhaustedPolicy: PoolExhaustedPolicy
     lock: Lock
   PoolExhaustedError* = object of CatchableError
@@ -15,7 +16,7 @@ type
     pepExtend
 
 
-func newPool*(defaultSize: Positive, poolExhaustedPolicy = pepRaise): Pool =
+func newPool*[T: Conn](defaultSize: Positive, poolExhaustedPolicy = pepRaise): Pool[T] =
   ##[ Create a connection pool of the given size.
 
   ``poolExhaustedPolicy`` defines how the pool reacts when a connection is requested but the pool has no connection available:
@@ -23,12 +24,17 @@ func newPool*(defaultSize: Positive, poolExhaustedPolicy = pepRaise): Pool =
     - ``pepRaise`` (default) means throw ``PoolExhaustedError``
     - ``pepExtend`` means throw “add another connection to the pool.”
   ]##
-  result = Pool(defaultSize: defaultSize, conns: newSeq[DbConn](defaultSize), poolExhaustedPolicy: poolExhaustedPolicy)
+
+  result = Pool[T](defaultSize: defaultSize, conns: newSeq[T](defaultSize), poolExhaustedPolicy: poolExhaustedPolicy)
 
   initLock(result.lock)
 
   for conn in result.conns.mitems:
-    conn = getDb()
+    conn =
+      when T is sqlite.DbConn:
+        sqlite.getDb()
+      elif T is postgres.DbConn:
+        postgres.getDb()
 
 func defaultSize*(pool: Pool): Natural =
   pool.defaultSize
@@ -36,10 +42,10 @@ func defaultSize*(pool: Pool): Natural =
 func size*(pool: Pool): Natural =
   len(pool.conns)
 
-func pop*(pool: var Pool): DbConn =
+func pop*[T: Conn](pool: var Pool[T]): T =
   ##[ Take a connection from the pool.
 
-  If you're calling this manually, don't forget to `add <#add>`_ it back!
+  If you're calling this manually, don't forget to `add <#add,Pool,DbConn>`_ it back!
   ]##
 
   withLock(pool.lock):
@@ -50,9 +56,13 @@ func pop*(pool: var Pool): DbConn =
       of pepRaise:
         raise newException(PoolExhaustedError, "Pool exhausted")
       of pepExtend:
-        result = getDb()
+        result =
+          when T is sqlite.DbConn:
+            sqlite.getDb()
+          elif T is postgres.DbConn:
+            postgres.getDb()
 
-func add*(pool: var Pool, dbConn: DbConn) =
+func add*[T: Conn](pool: var Pool, dbConn: T) =
   ##[ Add a connection to the pool.
 
   Use to return a borrowed connection to the pool.
--- a/src/norm/pool.nim
+++ b/src/norm/pool.nim
@@ -9,6 +9,7 @@ type
     defaultSize: Natural
     conns: seq[T]
     poolExhaustedPolicy: PoolExhaustedPolicy
+    getDbProc: proc: T
     lock: Lock
   PoolExhaustedError* = object of CatchableError
   PoolExhaustedPolicy* = enum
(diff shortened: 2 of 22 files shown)


tests/config.nims
```diff
@@ -1,3 +1,4 @@
 switch("path", "$projectDir/../src")
 switch("define", "normDebug")
 switch("deepcopy", "on")
+switch("mm", "orc")
```
moigagoo, line 4: @PhilippMDoerner interestingly enough, I couldn't make the tests run with the default memory manager. I would get a nil access error when closing the pool after extending it. No idea why that happened but I'm more or less OK to require ORC for Norm.

moigagoo:
@PhilippMDoerner please take a look at this PR. I've only implemented Pool for SQLite so far, just to test-drive the concept.

So far, the concept looks viable. I've tested it against concurrent access, seems to work.

Two policies for pool exchaustion were implemented: raise and extend. You can reset the pool after extension.

tests/config.nims
PhilippMDoerner: Agreed on the basis that ORC is the default for nim 2.0 anyway

src/norm/pool.nim
```diff
@@ -0,0 +1,69 @@
...
+    conns: seq[DbConn]
+    poolExhaustedPolicy: PoolExhaustedPolicy
+    lock: Lock
+  PoolExhaustedError* = object of CatchableError
+  PoolExhaustedPolicy* = enum
+    pepRaise
+    pepExtend
+
+
+func newPool*(defaultSize: Positive, poolExhaustedPolicy = pepRaise): Pool =
```
PhilippMDoerner, line 18: I like that this goes with nims default naming convention of "new" for ref objects and "init" for value objects!
Good catch, totally overlooked that in tinypool

src/norm/pool.nim
```diff
@@ -0,0 +1,69 @@
...
+      close conn
+
+func close*(pool: var Pool) =
+  withLock(pool.lock):
+    for conn in pool.conns.mitems:
+      close conn
+
+    pool.conns.setLen(0)
+
+template withDb*(pool: var Pool, body: untyped): untyped =
```
PhilippMDoerner, line 81: I feel tempted to allow the user to define the variable name simply because its "less magic" in a sense.
Would extend the template slightly, but make it more behave how I'd anticipate:
```nim
template withDb*(pool: var Pool, db: untyped, body: untyped) =
  block:
    var db = pool.pop()
    try:
      body
    finally:
      recycleConnection(connection)
  ```

Then again, the current way is more consistent with norms previous API for this.

Comment (moigagoo) [reply src/norm/pool.nim:81]: > Then again, the current way is more consistent with norms previous API for this.

Exactly. I've just taken the existing ``withDb`` and added a pool to it. So that it takes its ``db`` variable value from the pool and not from the environment.

I think we can add this other flavor of ``withDb`` as well, it's just a whole different story.

**Comment by PhilippMDoerner** [comment]
Do you have any plans yet on if you want to have a shared module with all key procs / if you want to make this module generic?

**Comment by moigagoo** [comment]
@PhilippMDoerner 

> Do you have any plans yet on if you want to have a shared module with all key procs / if you want to make this module generic?
The only thing I can see so far is the differing implementations on getDb between the sqlite and the postgres module. Feels tempting to have a single generic pool module, import and export that in both sqlite/postgres and use mixin getDb wherever getDb is used.

I would prefer a single module but so far I can't see how I can do that. If I have a wrapper type `DbConn = sqlite.DbConn | postgres.DbConn`, I get these “ambigous call” errors because it's now unclear what `seq[DbConn]` or `close conn` means. The compiler can't decide which DbConn I mean here.

**Comment by PhilippMDoerner** [comment]
I'll need to play around with that later when I can.
Worst case scenario I'd suggest a core pool modules + one for sqlite/postgres who do nothing but call the core modules but now with explicit types.
The reason why I'm suggesting that is because it centralises business logic in one place. It does mean annoying glue code, but I'd prefer that over duplicated business logic

15.10.2022 12:16:09 Constantine Molchanov ***@***.***>:

**Comment by moigagoo** [comment]
@PhilippMDoerner sure, I hate copy-pasted code too. I've tried using generics, to no avail so far. I'll give it some more thought.

Do I have your approval on the concept in general?

File: src/norm/pool.nim [hunk 4]
```diff
@@ -0,0 +1,106 @@
...
+    - ``pepExtend`` means throw “add another connection to the pool.”
+  ]##
+
+  result = Pool[T](defaultSize: defaultSize, conns: newSeq[T](defaultSize), poolExhaustedPolicy: poolExhaustedPolicy)
+
+  initLock(result.lock)
+
+  for conn in result.conns.mitems:
+    conn =
+      when T is sqlite.DbConn:
```
moigagoo, line 34: @PhilippMDoerner I've managed to make the implementation universal by making Pool type generic and using this kind of ugly hack. I'm not super happy with this part but it does the job done and do so without any runtime penalty.

PhilippMDoerner: Given that you do this in two different places it might be worth considering moving this "when"-ified getDb call into a private proc that deals with this

src/norm/private/postgres/rowutils.nim
```diff
@@ -21,20 +21,20 @@ func isEmptyColumn*(row: Row, index: int): bool =
 
 ## This does the actual heavy lifting for parsing
 proc fromRowPos[T: ref object](obj: var T, row: Row, pos: var Natural, skip: static bool = false) =
-  ##[ Convert ``ndb.sqlite.Row`` instance into `ref object`_ instance, from a given position.
```
PhilippMDoerner, line 24: I'd disagree with this doc comment change, as it is no longer solely accurate. With the move to ref object in order to allow rawSelect, this does more than just convert Models

src/norm/pool.nim
```diff
@@ -0,0 +1,106 @@
...
+      elif T is postgres.DbConn:
+        postgres.getDb()
+
+func defaultSize*(pool: Pool): Natural =
+  pool.defaultSize
+
+func size*(pool: Pool): Natural =
+  len(pool.conns)
+
+func pop*[T: DbConn](pool: var Pool[T]): T =
```
PhilippMDoerner, line 45: A log of level debug could be useful here to log the that after taking out/before taking out there's X connections. Though arguably the user could also do this themselves if needed before/after calling `pop`. Same for `add` and maybe `reset`

PhilippMDoerner (approved):
Overall looks good to me. I'm personally a fan of the "borrow/recycle" naming since "borrow" indicates that you're supposed to return the connection, but "pop/add" suits the general idea that the user can do whatever they want with the pool, you're just providing some utility procs to deal with it.

Though I do disagree with the doc comment changes from ref object to Model as stated in the comment for it.

PhilippMDoerner:
Sidenote!
It could make sense to add a "warn" log or the like to the `add` proc if the amount of connections in there exceeds a specified multiple of the pools default size.
If, for example, the pool contains 1000x the amount of connections that the pool was originally sized for, I'd assume something is fishy.

changelog.md
```diff
@@ -8,13 +8,19 @@
 -   [d]—docs improvement
 
 
+## 2.5.3 (WIP)
```
PhilippMDoerner, line 11: I would argue this is a minor version change, so 2.6.0

PhilippMDoerner:
Actually, wait!
We might want to contemplate **not** using getDb().
Why is that? A significant benefit of allowing multiple, user defined pools is the fact that you might want to connect to multiple databases.
If we use getDb, that is not the case.
We might want to contemplate allowing the user to pass a proc that generates DbConn instead and use that instead of getDb!

src/norm/pool.nim
PhilippMDoerner: Actually, wait!
We might want to contemplate **not** using getDb() at all.

A significant benefit of allowing multiple, user defined pools is the fact that you might want to connect to multiple databases.
If we use getDb, that is not the case, because you'll always be using the same environment variables.

We might want to contemplate allowing the user to pass a proc that generates DbConn instead and use that instead of getDb! getDb could be the default proc for such a parameter though.

moigagoo:
> It could make sense to add a "warn" log or the like to the add proc if the amount of connections in there exceeds a specified multiple of the pools default size.
If, for example, the pool contains 1000x the amount of connections that the pool was originally sized for, I'd assume something is fishy.

Sounds like an interesting idea! Maybe make this number configurable with do:dangerousPoolSize=1000 with a default of 100?

PhilippMDoerner:
> > It could make sense to add a "warn" log or the like to the add proc if the amount of connections in there exceeds a specified multiple of the pools default size.
> > If, for example, the pool contains 1000x the amount of connections that the pool was originally sized for, I'd assume something is fishy.
> 
> Sounds like an interesting idea! Maybe make this number configurable with -d:normdPoolWarning=1000 with a default of 100?

Works for me, that way only folks that want this kind of warning can opt into it, everybody else can just ignore it.

moigagoo:
> 'm personally a fan of the "borrow/recycle" naming since "borrow" indicates that you're supposed to return the connection, but "pop/add" suits the general idea that the user can do whatever they want with the pool, you're just providing some utility procs to deal with it.

I started with borrowDb and returned and ended up with pop and add because I suddenly realised that:
1. Theoretically, you can add a connection that has never been in the pool. So it's not necessarily returning, it may be just adding.
2. Pop and add are primitives just like pop and add for a seq. So it may be a good idea to name them similarly to their seq counterparts.

src/norm/pool.nim
```diff
@@ -0,0 +1,99 @@
...
+    poolExhaustedPolicy: PoolExhaustedPolicy
+    getDbProc: proc: T
+    lock: Lock
+  PoolExhaustedError* = object of CatchableError
+  PoolExhaustedPolicy* = enum
+    pepRaise
+    pepExtend
+
+
+proc newPool*[T: DbConn](defaultSize: Positive, poolExhaustedPolicy = pepRaise, getDbProc: proc: T = getDb): Pool[T] =
```
PhilippMDoerner, line 20: Does that actually get past the tests?
I'm a bit surprised, I'd have assumed that the compiler starts nagging about how it can't figure out whether to use sqlite.DbConn or postgres.DbConn

src/norm/pool.nim
PhilippMDoerner: Just for reference, I consider my comment here as resolved as you implemented it as the user passing in a connection-creation proc.

src/norm/pool.nim
PhilippMDoerner: Given that the tests ran through, it actually does work... I'm so confused. Oh well, nice to see it function!

moigagoo:
@PhilippMDoerner OK I'm ready to merge. I've decided not to add logging for now as this is would slow me down and this can be safely added later. This has little to do with pooling per se, it's more about adding a compilation flag and modifying the logging proc and then putting it all together.

If you don't see any blockers, I'll merge the PR.

PhilippMDoerner:
Looks good! Really happy to see norm have connection pooling!

PR #181, opened 2022-12-10
Implement indexes.

moigagoo:


moigagoo:
Nimibook would not build. Here is the same issue in SciNim: https://github.com/SciNim/getting-started/pull/50

Issue #184, opened 2023-01-22
[Bug?] SQLITE-Foreignkeys pragma is off by default whenever a new connection is opened and acts on a per connection basis


PhilippMDoerner:
What I'm about to describe is an issue with **sqlite***, not with norm itself. 
Norm could however possibly address what I deem unexpected behaviour.
I just noticed this one because I found a bug in my database that should've been quite literally impossible.

<h3> The Problem </h4>
Here a reproduction with a minimal example:

```sql
DROP TABLE IF EXISTS "campaign";
DROP TABLE IF EXISTS "creature";

CREATE TABLE "campaign" (
    "id" integer NOT NULL PRIMARY KEY AUTOINCREMENT,
     "name" varchar(200) NOT NULL UNIQUE
);
CREATE TABLE "creature" (
    "id" integer NOT NULL PRIMARY KEY AUTOINCREMENT,
    "name" varchar(200) NOT NULL UNIQUE,
    "campaign_id" REFERENCES "campaign" ("id") DEFERRABLE INITIALLY DEFERRED
);

INSERT INTO "campaign" VALUES (1, "MyCampaign");
INSERT INTO "creature" VALUES (1, "cre1", 1), (2, "cre2", 1);

DELETE FROM "campaign" WHERE id = 1;
```

If you run this, you will NOT get an error, even though you're deleting the entry "MyCampaign" which is referenced by 2 other entries!
That is because foreign key checks are not turned on by default, that is done via the PRAGMA statement.

`PRAGMA foreign_keys=on;`

So by default norm (because of the way sqlite behaves) does not do foreign key checks before deleting, inserting or updating an entry.
Now you might **think** that this just needs to be done once on the db file and then it works in perpetuity,

[Think again](https://sqlite.org/forum/info/c5dc50f61b88c587). That PRAGMA needs to be turned on on a **per connection** basis. Here a minimal example using db_sqlite to prove this (though you can also try this out with the sqlite3 binary by connecting to a dbfile, enabling the pragma, disconnecting and then connecting again, it displays the exact same behaviour):

```nim
import std/db_sqlite 
let db1 = open("fk.sqlite3", "", "", "")
let db2 = open("fk.sqlite3", "", "", "")

proc printPragma(con: DbConn) =
  echo con.getRow(sql"PRAGMA foreign_keys;")

db1.printPragma() # @["0"] <-- Pragma is off
db1.exec(sql"PRAGMA foreign_keys=on;")
[...]

PR #185, opened 2023-01-25
Fix sqlite pragma issue

PhilippMDoerner:
Implements #184 in the following scope:

- Enables Foreign key Pragma for all connections created via getDb()
- Modifies "getDbProc" when creating an sqlite Pool to also enable the Foreign Key Pragma
- Adds docs educating about the situation in case the user wants to have that behaviour on a connection they created themselves
- Adds tests checking that a DB Error correctly occurs if a Foreign-key Constraint Violation occurs
- Updated changelog
- Added equal tests to Postgres so the tests are symmetrical and we're making sure that both cover the exact same scope of behaviour


moigagoo:
@PhilippMDoerner I'm happy with the PR but there's just one thing I'd like you to change. It's something so minor I'm feeling awkward even bringing this up. But could you please spell “Norm” and not “norm” in your copy? :-)

PhilippMDoerner:
No worries. By "in your copy" you mean in the book docs? Or just generally go over the project and replace `Norm` with `norm` where its not at the start of a sentence?

moigagoo:
> No worries. By "in your copy" you mean in the book docs? Or just generally go over the project and replace Norm with norm where its not at the start of a sentence?

I meant all places where it's not code. I.e., docs, book, comments.

PhilippMDoerner:
Done!
I did leave "Norm" in capital letter at the start of sentences for grammatical correctness there (at least IIRC if it's at the start of a sentence it's always capitalized, no exceptions). So there's still sentences like this:
`Norm's `getDb` proc lets you create a DB connection using `DB_HOST`, `DB_USER`, `DB_PASS`, and `DB_NAME` environment variables:`

If you want me to change those as well, give a shout, other than that the most recent commit should've taken care of it.

moigagoo:
@PhilippMDoerner https://github.com/moigagoo/norm/pull/185/commits/fe4f660026a2624459b681696c00d85d810bad06 It's the other way around :-)

Capitalized “Norm” is the right spelling. I've noticed that you tend to use “norm” instead.

PhilippMDoerner:
Wow I'm tired xD

Alrighty, one sec

PhilippMDoerner:
Okay, here we go, attempt number 2 of 2 possible solutions

PR #186, opened 2023-02-25
Nim2 compatibility

### LINKED ISSUE #182: Add compatibility for nim 2.0
I know you're already aware of this, but I believe making it an issue would still be useful to document our trail of thoughts as well as progress.

With nim 2.0 the packages std/db_sqlite/db_postgres/db_mysql are no longer part of the standard library!

They have been moved into its own package called [dbconnector](https://github.com/nim-lang/db_connector).
I do not believe this affects norm directly as I'm not aware of any std/db_* imports there, but it does affect norm's ndb dependency.
It makes use of std/sqlite3 and stb/db_common.

Over there it should be as simple as swapping out the imports:

    std/sqlite3 => db_connector/sqlite3
    std/db_common => db_connector/db_common

Once this is fixed, it will require an update of the ndb dependency.

Now the question is what to do about this?
ndb appears to not be receiving any further work.
Should we fork?
Or PR to the original ndb upstream?
Or remove the ndb dependency entirely and move that code into norm?

The maintainer of ndb does appear to still accept PRs, so working with that repo seems still possible.
I personally would prefer not to move ndb code into norm, as the value of a postgres/sqlite wrapper that allows for more correctness is something not just norm can make use of.

PhilippMDoerner:
closes #182 

Makes norm viable with nim devel:
- Swaps ndb for lowdb, which is useable with the current version of nim-devel
- Swaps out generics using `[_]` to `[T]` etc. due to breaking change introduced by https://github.com/nim-lang/Nim/pull/21192
See: https://github.com/nim-lang/Nim/issues/21435

Unrelated to that:
I introduced nimble tasks to run the tests because I got tired of looking up the exact command after the third time and I'd prefer just having a nimble task for everything related to a package :sweat_smile:

### DIFF
--- a/norm.nimble
+++ b/norm.nimble
@@ -10,7 +10,7 @@ skipDirs      = @["tests", "htmldocs"]
 
 # Dependencies
 
-requires "nim >= 1.4.0", "ndb >= 0.19.9"
+requires "nim >= 1.4.0", "lowdb >= 0.1.0"
 
 
 task test, "Run tests":
@@ -19,8 +19,8 @@ task test, "Run tests":
 task book, "Generate book":
   rmDir "docs"
   exec "nimble install -y nimib nimibook@#280a626a902745b378cc2186374f14c904c9a606"
-  exec "nim r -d:release nbook.nim update"
-  exec "nim r -d:release nbook.nim build"
+  exec "nim r -d:release --mm:refc nbook.nim update"
+  exec "nim r -d:release --mm:refc nbook.nim build"
   cpFile("CNAME", "docs/CNAME")
 
 task docs, "Generate docs":
--- a/norm.nimble
+++ b/norm.nimble
@@ -10,7 +10,7 @@ skipDirs      = @["tests", "htmldocs"]
 
 # Dependencies
 
-requires "nim >= 1.4.0", "lowdb >= 0.1.0"
+requires "nim >= 1.4.0", "lowdb >= 0.1.1"
 
 
 task test, "Run tests":
--- a/norm.nimble
+++ b/norm.nimble
@@ -27,24 +27,39 @@ task docs, "Generate docs":
   rmDir "docs/apidocs"
   exec "nimble doc --outdir:docs/apidocs --project --index:on src/norm"
 
+
+## For Local Development
+
 import std/[strutils, sequtils, strformat]
 
-task dockerTests, "Run all containerized tests":
-  let wantsSudoExecution = commandLineParams.anyIt(it == "sudo")
-  var command = "docker-compose run --rm tests"
-  if wantsSudoExecution:
+let postgresName = "norm-postgres-testcontainer"
+putEnv("PGHOST", "localhost") ## Mandatory for all Postgres tests
+
+proc asSudo(params: seq[string]): bool =
+  return params.anyIt(it == "sudo")
+
+task startContainers, "Starts a postgres container for running tests against":
+  var command = fmt"""docker run -d -e POSTGRES_PASSWORD="postgres" --name {postgresName} --rm -p 5432:5432 postgres"""
+
+  if commandLineParams.asSudo():
     command = fmt"sudo {command}"
 
   exec command
 
-task singleDockerTest, "Run containerized tests for a specific test file":
-  let testArguments = commandLineParams.filterIt(it.startsWith("test"))
+task stopContainers, "Stops a postgres container used for norm tests":
+  var command = fmt"""docker stop {postgresName}"""
 
-  let wantsSudoExecution = commandLineParams.anyIt(it == "sudo")
-  var command = "docker-compose run --rm tests"
+  if commandLineParams.asSudo():
+    command = fmt"sudo {command}"
 
-  for argument in testArguments:
-    var command = fmt"docker-compose run --rm test {argument}"
-    if wantsSudoExecution:
-      command = fmt"sudo {command}"
+  exec command
+
+task allTests, "Run all tests via testament":
+  exec "testament all"
+
+task singleTest, "Run containerized tests for a specific test file":
+  let testFiles = commandLineParams.filterIt(it.startsWith("test"))
+
+  for file in testFiles:
+    let command = fmt"nimble c -r {file}"
     exec command
--- a/norm.nimble
+++ b/norm.nimble
@@ -18,7 +18,7 @@ task test, "Run tests":
 
 task setupBook, "Compiles the nimibook CLI-binary used for generating the docs":
   exec "nimble install -y nimib nimibook@#280a626a902745b378cc2186374f14c904c9a606"
-  exec "nim c -d:release --mm:orc --deepcopy:on nbook.nim"
+  exec "nim c -d:release --mm:refc nbook.nim"
 
 task book, "Generate book":
   rmDir "docs"
--- a/norm.nimble
+++ b/norm.nimble
@@ -16,11 +16,15 @@ requires "nim >= 1.4.0", "lowdb >= 0.1.1"
 task test, "Run tests":
   exec "testament all"
 
+task setupBook, "Compiles the nimibook CLI-binary used for generating the docs":
+  exec "nimble install -y nimib nimibook@#280a626a902745b378cc2186374f14c904c9a606"
+  exec "nim c -d:release --mm:orc --deepcopy:on nbook.nim"
+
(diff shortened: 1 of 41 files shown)


src/norm/private/sqlite/dbtypes.nim
```diff
@@ -90,9 +90,9 @@ func to*(dbVal; T: typedesc[DbBlob]): T = dbVal.b
 
 proc to*(dbVal; T: typedesc[DateTime]): T = utc dbVal.f.fromUnixFloat()
 
-func to*[_](dbVal; T: typedesc[StringOfCap[_]]): T = dbVal.s.T
+func to*[U](dbVal; T: typedesc[StringOfCap[U]]): T = dbVal.s.T
 
-func to*[_](dbVal; T: typedesc[PaddedStringOfCap[_]]): T = dbVal.s.T
+func to*[U](dbVal; T: typedesc[PaddedStringOfCap[U]]): T = dbVal.s.T
```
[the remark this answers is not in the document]
PhilippMDoerner: It just had to be something that wasn't "T", as "T" was already taken for the parameter name.

norm.nimble
```diff
@@ -19,11 +19,32 @@ task test, "Run tests":
...
+  if wantsSudoExecution:
+    command = fmt"sudo {command}"
+
+  exec command
+
+task singleDockerTest, "Run containerized tests for a specific test file":
+  let testArguments = commandLineParams.filterIt(it.startsWith("test"))
+
+  let wantsSudoExecution = commandLineParams.anyIt(it == "sudo")
+  var command = "docker-compose run --rm tests"
```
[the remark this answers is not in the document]
PhilippMDoerner: Want me to make it a proc that acts on commandLineParams to make the nimble tasks slightly cleaner?

PhilippMDoerner:
Okay, here we go. @moigagoo I had to significantly revamp the test-suite, solely so we can start testing multiple nim-versions there instead of just one. Couldn't figure out a way that left the calling code within a container itself.

As was visible with the dbTypes issue, it was quite necessary to start testing for multiple versions.

That did sacrifice the docker-compose commands, but I replaced them with (hopefully) as convenient nimble tasks.

Anyway:
- Testsuite runs in github :heavy_check_mark: 
- Testsuite runs locally :heavy_check_mark: 
- Testuite expanded to run once for 1.6.10 (current stable) and devel (future nim 2.0) :heavy_check_mark: 
- Provided new set of commands to easily execute testsuite :heavy_check_mark:

norm.nimble
```diff
@@ -19,11 +19,32 @@ task test, "Run tests":
 task book, "Generate book":
   rmDir "docs"
   exec "nimble install -y nimib nimibook@#280a626a902745b378cc2186374f14c904c9a606"
-  exec "nim r -d:release nbook.nim update"
-  exec "nim r -d:release nbook.nim build"
+  exec "nim r -d:release --mm:refc nbook.nim update"
```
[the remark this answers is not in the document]
PhilippMDoerner: The answer to that is yes.
Just updated the nimble task and compiled it locally, works fine.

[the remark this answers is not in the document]
PhilippMDoerner: I stand corrected.
The answer is yes, but only on nim devel.
Currently the book is still compiled with nim 1.6.10, which has a critical bug at one point that makes it impossible compile the book.

src/norm/postgres.nim
```diff
@@ -217,7 +217,7 @@ proc insert*[T: Model](dbConn; objs: var openArray[T], force = false) =
   for obj in objs.mitems:
     dbConn.insert(obj, force)
 
-proc select*[T: Model](dbConn; obj: var T, cond: string, params: varargs[DbValue, dbValue]) {.raises: {NotFoundError, ValueError, DbError, LoggingError}.} =
+proc select*[T: Model](dbConn; obj: var T, cond: string, params: varargs[DbValue, dbValue]) {.raises: {NotFoundError, ValueError, DbError, LoggingError, Exception}.} =
```
moigagoo, line 220: What is the point of this generic exception? Doesn't it include all possible exceptions?

src/norm/postgres.nim
```diff
@@ -433,7 +433,7 @@ template transaction*(dbConn; body: untyped): untyped =
     log(commitQry)
     dbConn.exec(sql commitQry)
 
-  except:
+  except CatchableError:
```
moigagoo, line 436: Does CatchableError cover all cases? You want your transactions rolled back if anything bad happens.

src/norm/private/postgres/dbtypes.nim
```diff
@@ -21,7 +21,7 @@ func dbType*(T: typedesc[int16]): string = "SMALLINT"
 
 func dbType*(T: typedesc[int32]): string = "INTEGER"
 
-func dbType*(T: typedesc[int64]): string = "BIGINT"
+func dbType*(T: typedesc[int64 | Positive | int]): string = "BIGINT"
```
moigagoo, line 24: If Positive is here then Natural should be too?

src/norm/private/postgres/dbtypes.nim
```diff
@@ -67,9 +67,9 @@ func to*(dbVal; T: typedesc[SomeFloat]): T = dbVal.f.T
 
 func to*(dbVal; T: typedesc[string]): T = dbVal.s
 
-func to*[_](dbVal; T: typedesc[StringOfCap[_]]): T = dbVal.o.value.T
+func to*[U](dbVal; T: typedesc[StringOfCap[U]]): T = dbVal.o.value.T
 
-func to*[_](dbVal; T: typedesc[PaddedStringOfCap[_]]): T = dbVal.o.value.T
+func to*[U](dbVal; T: typedesc[PaddedStringOfCap[U]]): T = dbVal.o.value.T
```
moigagoo, line 72: I think U is confusing since it doesn't mean anything.

How about T1 and T2?

src/norm/private/sqlite/dbtypes.nim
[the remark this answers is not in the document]
moigagoo: I suggest T1 and T2.

moigagoo (commented):
I have a few minor comments but in general I'm happy with this PR.

Thank you so much 🙏

src/norm/postgres.nim
PhilippMDoerner: Nim appears to have gotten a damn sight tighter about its exception tracking system.
You'll notice that `log` also raises Exception. That's because the procs `log` calls (`logLoop` from logging.nim) can raise `Exception`. That in turn means that `select`, which calls `log`, can raise `Exception`. Therefore, in order for this to compile, we **must** add `Exception` there. Same for all the other places as well as `log.nim`.

src/norm/postgres.nim
PhilippMDoerner: It catches any error you should be able to catch. It does not catch defects, but with panics:on, which is what you should be compiling with as that is how they are designed, defects are uncatchable. They are the "root"-error type, if you will.

DBError: CatchableError --> IOError --> DBError
OSError: CatchableError --> OSError

src/norm/private/postgres/dbtypes.nim
PhilippMDoerner: We *can* do this, but we don't have to. If we add it, it'd be to be explicit, not because the code requires it. I'd leave that up to you.

We can leave it out, as this still works:

```nim
func dbType*(T: typedesc[int64 | Positive | int ]): string = "BIGINT"
let x: Natural = 5
echo dbType(x.typeOf)
```

The reason for that is `int`, it can act as a stand-in for `Natural`, so nim counts this as "valid". 
If you want to be explicit about it, I can add it in.

src/norm/private/postgres/dbtypes.nim
PhilippMDoerner: Works for me, Done

src/norm/postgres.nim
ringabout: I don't think it is the right way. You might somehow remove `Exception` from the raises lists of callees.

ringabout: For example, you might put `log` under the `try` statements

```nim
try:
  log ...
except Exception:
  raises newexception(LoggingError, ...)
```

moigagoo (approved):

PR #187, opened 2023-03-12
Update lowdb version to 0.2.0 and add enum support

### LINKED ISSUE #166: [Feature Request] Enum Support
You've already indirectly talked about enum support in
https://github.com/moigagoo/norm/issues/34#issuecomment-544385989

Enums feel like a rather important thing in nim to me, so I think it would great to support it out of the box.

Since Postgres has an enumtype it should be kinda straight forward.

Regarding your question how to handle enums in sqlite:
I have this in my project atm, which encodes enums as strings and parses them back using `std/strutils.parseEnum`

```nim
func dbType*(T: typedesc[enum]): string = "STRING"
func dbValue*(val: enum): DbValue = DbValue(kind: dvkString, s: $val)
func to*(dbVal: DbValue, T: typedesc[enum]): T = parseEnum[T](dbVal.s)
```

I imagine this could be rather costly for huge databases, so I don't know if it's a great solution.
However the benefit is, that the meaning of the value is preserved in the database.

The alternative would be to encode the enum as ints, obviously. Probably that is the better default choice.
The best way would be, to enable norm users to choose I guess.

Thanks for all the work put into norm. It's great!

PhilippMDoerner:
Lowdb is used for direct interaction with the database. This means easier enum support.
For supporting any new datatype, 3 procs are needed:
- to
- dbType
- dbValue

In the newest version it provides propper enum support, meaning no dbValue proc needs to be defined for it. Norm only needs to define "to" and "dbType".
These procs have now been added.

Further, tests were added to confirm that these procs can be overwritten by procs specific for a given enum.

Note:
I will be honest, I don't have the faintest clue where these types of tests "properly" belong. So I *think* they're correctly placed where they are (?).

Closes #166

### DIFF
--- a/norm.nimble
+++ b/norm.nimble
@@ -1,6 +1,6 @@
 # Package
 
-version       = "2.7.1"
+version       = "2.7.0"
 author        = "Constantine Molchanov"
 description   = "Nim ORM for SQLite and PostgreSQL."
 license       = "MIT"
--- a/norm.nimble
+++ b/norm.nimble
@@ -1,6 +1,6 @@
 # Package
 
-version       = "2.7.0"
+version       = "2.7.1"
 author        = "Constantine Molchanov"
 description   = "Nim ORM for SQLite and PostgreSQL."
 license       = "MIT"
@@ -10,7 +10,7 @@ skipDirs      = @["tests", "htmldocs"]
 
 # Dependencies
 
-requires "nim >= 1.4.0", "lowdb >= 0.1.1"
+requires "nim >= 1.4.0", "lowdb >= 0.2.0"
 
 
 task test, "Run tests":
--- a/src/norm/private/postgres/dbtypes.nim
+++ b/src/norm/private/postgres/dbtypes.nim
@@ -21,7 +21,7 @@ func dbType*(T: typedesc[int16]): string = "SMALLINT"
 
 func dbType*(T: typedesc[int32]): string = "INTEGER"
 
-func dbType*(T: typedesc[int64 | Positive | int | Natural]): string = "BIGINT"
+func dbType*(T: typedesc[int64 | Positive | int | Natural | enum]): string = "BIGINT"
 
 func dbType*(T: typedesc[float32]): string = "REAL"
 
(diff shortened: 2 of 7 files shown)


norm.nimble
```diff
@@ -1,6 +1,6 @@
 # Package
 
-version       = "2.7.0"
+version       = "2.7.1"
```
moigagoo, line 3: @PhilippMDoerner don't bump yet please. We bump on release.

src/norm/private/postgres/dbtypes.nim
```diff
@@ -21,7 +21,7 @@ func dbType*(T: typedesc[int16]): string = "SMALLINT"
 
 func dbType*(T: typedesc[int32]): string = "INTEGER"
 
-func dbType*(T: typedesc[int64 | Positive | int | Natural]): string = "BIGINT"
+func dbType*(T: typedesc[int64 | Positive | int | Natural | enum]): string = "BIGINT"
```
moigagoo, line 24: @PhilippMDoerner I wonder if it's more appropriate to store enum as INTEGER and not BIGINT. You probably won't have a BIGINT number of enum values.

tests/models.nim
```diff
@@ -88,6 +88,18 @@ type
     lastName* {.index: "idx_student_names".}: string
     email* {.uniqueIndex: "idx_student_emails".}: string
 
+  CoinKind* = enum
+    ck1Cent, ck2Cents, ck5Cents, ck10Cents, ck20Cents, ck50Cents ck1Euro, ck2Euro
```
moigagoo, line 92: Is there a comma missing?

moigagoo (changes requested):

src/norm/private/postgres/dbtypes.nim
PhilippMDoerner: Fair point, I agree with the reasoning. I changed it to use INTEGER instead. If somebody needs something else, they can just write their own dbType as that seems to me to be a special case.

moigagoo (approved):

PR #194, opened 2023-09-08
Moar examples

### LINKED ISSUE #193: [Documentation Request] Add examples for using more fancy SQL such as `ORDER BY`, `LIMIT` etc. with `select`
System64 - Flandre Scarlet over at the nim-discord raised a good point about the docs:
There are no examples how to use the more fancy SQL elements that come after "WHERE" and we never make it clear you can use them with `select`.

That lead to them using `rawSelect`, which is **not** the intended usecase. We should add examples for those docs.
HOWEVER, as the `Rows` Section is already pretty large, I'd like to split the `Rows` section up into 3 pages:
- Simple SQL Syntax + Examples to deal with Rows
- Fancy SQL Syntax + Examples to deal with Rows (order by, Limit etc.)
- Selecting Many-to-One/One-to-Many/Many-to-Many relationships

Do I have consensus for that from you @moigagoo?
Just to elaborate on why I'm not just doing it: I just don't want to surprise you ;-)
Simple response or thumbsup/down emoji would suffice.

PhilippMDoerner:
closes #193 

Sidenote: It *would* be mildly useful if I could actually see the consequences of this nimibook update somewhere =/

### DIFF
--- a/book/tutorial/rows.nim
+++ b/book/tutorial/rows.nim
@@ -136,107 +136,50 @@ nbCode:
   for pet in petsFoo:
     echo pet[]
 
-nbText: """
+nbText"""
+## Update Rows
 
-### Selecting Many-To-One/One-To-Many relationships
-
-Imagine you had a Many-To-One relationship between two models, like we have with `Customer` being the many-model and `User` being the one-model, where one user can have many customers. 
-
-If you have a user and wanted to query all of their customers, you couldn't do so by just making a query for the user, as that model doesn't have a "seq[Customer]" field that Norm could resolve.
-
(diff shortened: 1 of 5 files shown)


moigagoo (approved):
I've looked through the PR. I've spotted a bunch of minor editorial issues like wrong capitalization, missing commas, and so on.

However, I think all these are too minor to request any changes. Also, I'm infinitely grateful for you contributions and don't want to make it harder for you.

Therefore, I'm merging the PR as is. If I have time and strong enough motivation, I'll fix the spots that bother me later myself.

Thanks!

PR #199, opened 2023-11-22
Create Index If Not Exists

LichtBuch:
Added "IF NOT EXISTS" to index/uniqueIndex creation via pragmas. I did this so it's more in line with creating tables and to be able to run a migration regardless of whether the table is already built (before you would get an Exception when trying to build an existing database).


moigagoo:
@LichtBuch thanks for the contribution! Could you please add a changelog.md update to your PR? Just add a line about your change.

PR #209, opened 2025-01-24
Add back closure support

PhilippMDoerner:
Adds back closure support that was originally removed by this commit: https://github.com/moigagoo/norm/commit/7f766e4c12fa95392f67ca9b1dfef2fc6b7d2489

The "proc" annotation to the parameter type is *required* for this to support closures. If it isn't there, you only support essentially proc() = discard

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,123 +7,108 @@
 -   [t]—test suite improvement
 -   [d]—docs improvement
 
+## 2.8.6 (?? ??, 2025)
+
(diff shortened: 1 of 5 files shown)


PhilippMDoerner:
I noticed it for the most part because I updated and suddenly my compilation broke completely :smile:

moigagoo:
@PhilippMDoerner sorry for the trouble. I've noticed that the tests stopped passing with Nim 2.2.0 and fixed them 😅.

